PR #1, opened 2019-04-18
initial commit

dcasavant:
Initial commit of 8.2.1 Helm Charts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,26 @@
-# pega-k8s-deploy
+# Pega Deployment on Kubernetes
+
+This project provides Helm charts and basic examples for deploying Pega on Kubernetes. This project **does not include** the required database installation image which you [may obtain from the Pega Community](https://community.pega.com/knowledgebase/products/platform/deploy).
+
+
+
+## Getting Started
+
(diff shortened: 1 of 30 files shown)


cdancy:
We might want to rename this project to align more with what the broader community is doing. Looking at other projects, and think towards the future, we might want to rename this project to just `charts` and then layer this a directory or 2 deep. This would allow us a single place to store other charts that we may want to create in the future. Take a look at the below projects to get an idea of what I'm thinking about. 

https://github.com/helm/charts
https://github.com/bitnami/charts

dcasavant:
@cdancy Both of those examples include a number of charts - and we only have one for the time being.  I was looking more to application examples than chart libraries (but I am flexible!).  

I was also thinking that this repo would also hold the 'example yaml' files generated from these Helm charts.  I don't see that as a common pattern in other repos but it seems to have been well received so far by those who have had early access.  Maybe using a release is a better mechanism for that.

cdancy:
hmmm ... if the templates were well received than I wonder if it makes more sense to just port the entire build process over to github to allow folks to generate those templates themselves?

Or, and within the build process in mind, once we do a release we could absolutely wrap up the templates and publish them somewhere.

Issue #2, opened 2019-04-18
Add CI/CD Pipeline


dcasavant:
As a contributor to this repo, I would like to be able to validate my changes (styling/validation/lint/etc) to ensure my contributions are of the highest quality.

dcasavant:
Merged

PR #4, opened 2019-04-18
Issue/2

dcasavant:


dcasavant:
Fixing lint problems in the build - will be green now! #2

Issue #5, opened 2019-04-18
Generate example yaml


dcasavant:
The build should generate example yaml for all supported providers.

dcasavant:
Example yaml will be generated with a release is tagged.

dcasavant:
Generated https://github.com/pegasystems/pega-helm-charts/releases

PR #6, opened 2019-04-19
Update build process to generate example yaml

dcasavant:
Example yaml will be generated when a release is tagged.  This happens during the travis-ci build and then the archives are added to the release.  Additionally, it can be tested using _make examples_.


dcasavant:
#5

Issue #10, opened 2019-04-24
Add HPA for Web and Background tiers


dcasavant:
Incorporate an HPA into the deployment chart for the Web and Background tiers

dcasavant:
Merged!

cdancy:
Thanks @dcasavant !

Issue #3, opened 2019-04-18
Incorporate support for other cloud providers


dcasavant:
Incorporate the latest helm charts and migrate development to this repository.

dcasavant:
Add support for AKS, GKE, & PKS in the Helm charts

dcasavant:
Merged!

cdancy:
Thanks @dcasavant !

PR #11, opened 2019-04-26
Add latest changes for supporting GKE/PKS/AKS and HPAs

dcasavant:
Providing a number of enhancement to support additional Kubernetes providers as well as autoscaling of the web and background tiers.

Also see issues #10 and #3.


cdancy:
@dcasavant is the idea that we will just target latest internal code until we have a better idea surrounding how to release these?

dcasavant:
@cdancy These charts will always support the latest shipped version of Pega, but not unreleased work in progress.  The internal source for helm charts will also be replaced with this repository.

PR #26, opened 2019-10-14
Added Tier based deployment support, minikube support, Helm unit test cases, Auto-publish to bintray

MadhuriArugula:


charts/pega/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 name: pega
-version: "1.0.0"
+version: "0.0.1"
```
dcasavant: Why is this back to `0.0.1`?  Is it dynamically set as part of the deploy process so we don't need to worry about it here?

charts/pega/charts/installer/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
+name: installer
+version: 0.1.0
```
dcasavant: `0.1.0`?  Should this be brought up to be in sync with the rest of the Pega chart?

charts/pega/charts/installer/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+---
+apiVersion: v1
+appVersion: "1.0"
+description: A Helm chart for Kubernetes
```
dcasavant: If this description is put anywhere, we should make this more descriptive for the installer.

charts/pega/charts/installer/config/migrateSystem.properties.tmpl
```diff
@@ -0,0 +1,106 @@
+# Properties File for use with migrateSystem.xml  Update this file 
+# before using migrate.bat/sh script.
+# Set the DB connection
+
+################### COMMON PROPERTIES - DB CONNECTION (REQUIRED) ##################
```
dcasavant: This is not required to be set here since env variables are referenced.

charts/pega/charts/installer/config/prpcUtils.properties.tmpl
```diff
@@ -0,0 +1,86 @@
+# Properties file for use with PRPC Utilities. Update this file 
+# before using prpcUtils.bat/sh script.
+# For more information, see the Pega Platform help.
+
+################### COMMON PROPERTIES - DB CONNECTION (REQUIRED) ##################
```
dcasavant: Also probably not required since env variables are referenced?

charts/pega/charts/installer/config/setupDatabase.properties.tmpl
```diff
@@ -0,0 +1,143 @@
+# Properties file for use with Pega Deployment Utilities.  Update this file 
+# before using install/upgrade.bat/sh scripts.
+# For more information, see the Pega Platform help.
+
+################### COMMON PROPERTIES - DB CONNECTION (REQUIRED) ##################
```
dcasavant: Also probably not required since env variables are referenced?

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -0,0 +1,34 @@
...
+  
+    <Resource name="jdbc/PegaRULES"
+    auth="Container"
+    type="javax.sql.DataSource"
+    factory="org.apache.tomcat.jdbc.pool.DataSourceFactory"
+    driverClassName="{{ .Env.JDBC_CLASS }}"
+    url="{{ .Env.JDBC_URL }}"
+    username="{{ .Env.SECRET_DB_USERNAME }}"
+    password="{{ .Env.SECRET_DB_PASSWORD }}"
+    minActive="{{ .Env.JDBC_MIN_ACTIVE }}"
```
dcasavant: `minActive` isn't a thing in the jdbc spec as far as I can tell.

charts/pega/config/installer/migrateSystem.properties.tmpl
```diff
@@ -0,0 +1,106 @@
+# Properties File for use with migrateSystem.xml  Update this file 
+# before using migrate.bat/sh script.
+# Set the DB connection
+
+################### COMMON PROPERTIES - DB CONNECTION (REQUIRED) ##################
```
dcasavant: Also probably not required since env variables are referenced?

charts/pega/templates/_pega-service.tpl
```diff
@@ -16,12 +17,15 @@ metadata:
     traefik.ingress.kubernetes.io/max-conn-amount: '10'
     # Manually set the cookie name for sticky sessions
     traefik.ingress.kubernetes.io/session-cookie-name: UNIQUE-PEGA-COOKIE-NAME
```
dcasavant: Does this need to be a values.yaml parameter?  Maybe default the value to the deployment-tier name?

charts/pega/values-minimal.yaml
```diff
@@ -0,0 +1,89 @@
+---
+global:
```
dcasavant: Why are only some of the parameters under global? Is that correct? `docker` is in both.

terratest/src/test/Gopkg.lock
```diff
@@ -0,0 +1,695 @@
+# This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
```
dcasavant: Should this file be here or should it be removed? (Looks like a lock file that maybe should have been .gitignore'd)

kishorv10: I re-added the gopkg.lock file. The sporadic failures with the travis build  is observed after deleting it from git. It appears that Go uses this file for giving reproducible build when changes are moved across the git branches. Similar issue discussed here https://stackoverflow.com/questions/51738797/do-i-need-to-add-gopkg-lock-to-git
Also, checked in few other go repositories, gopkg.lock is preserved in the source code (quick ref: https://github.com/gruntwork-io/terratest).

charts/pega/Chart.yaml
kishorv10: This might be due to the generation of chart yaml during initial refactoring not in deploy process. As part of deploy process, Chart.yaml version gets auto updated only on the gz files uploaded to bintray. The version in the source code will remain same until some one manually changes it.
Shall we change it back to 1.0.0 or 1.2.0?

cdancy:
These PRs are getting bigger and bigger. Can't we break things up into smaller more targeted changes?

dcasavant:
@cdancy This should be the last big PR as it is merging the private repo with the open source repo.

Emilio-Pega:
Hi all, any ETA on when this is going to be merged/released? Octopodes **EPIC-55982** is going to BE BLOCKED by this after the current sprint...

cdancy:
> Hi all, any ETA on when this is going to be merged/released? Octopodes EPIC-55982 is going to BE BLOCKED by this after the current sprint...

Same for us :)

charts/pega/charts/installer/values.yaml
```diff
@@ -0,0 +1,54 @@
+---
+image: "YOUR_IMAGE_NAME"
+# Creates a new System and replaces this with default system.Default is pega
+systemName: "pega"
+# Creates the system with this production level.Default is 2
+productionLevel: 2
+# Whether this is a Multitenant System ('true' if yes, 'false' if no)
+multitenantSystem: "false"
+# UDF generation will be skipped if this property is set to true
+bypassUdfGeneration: "false"
```
taz-mon: should this be true by default??

Issue #28, opened 2019-11-13
Service configuration issue for AKS


taz-mon:
After I deployed, the exposed pega-web tier did not get an external IP address. It looks like the TYPE for the tier is NodePort, and for AKS at least, it should be LoadBalancer.
When I manually change pega-web type in the Pega service, the load-balancer provides an IP address immediately.
Should the service be configured so it is LoadBalancer for all environments and all exposed tiers?

Emilio-Pega:
If you go with LoadBalancer as the type, I'd still need the option for NodePort or some override of the "automatic" value

dcasavant:
Fix merged to master.

The service type will now default to `LoadBalancer`.  This is configurable per service in the tiers section.  See [the service documentation](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/README.md#service-optional) for more information.

Deployment Type | Load Balancer | Service Type
--- | --- | ---
AWS EKS | ALB | `LoadBalancer`
Minikube | None | `NodePort` (default in the minimal values file)
Other K8S environments | Traefik | `LoadBalancer`


This will be part of the v1.2.1 release.

Thanks for the submission and first cut at the fix @taz-mon !

PR #29, opened 2019-11-13
change default type of service from LoadBalancer

taz-mon:
changing from `NodePort` to `LoadBalancer` so the pega service gives IP Address to pega-web tiers

#28


Emilio-Pega:
Internally for SDEs we still want NodePort, PLEASE provide a way to allow NodePort to be set (via a value)

cdancy:
LGTM

Emilio-Pega:
LGTM

PR #32, opened 2019-11-25
Calculated web requestor passivation timeout based on alb_stickiness_…

nowam1:
Calculated web requestor passivation timeout based on alb_stickiness_lb_cookie_duration_seconds and requestorPassivationToAlbStickinessMargin.
Configured web requestor passivation timeout in prconfig based on calculated value.

### DIFF
--- a/charts/pega/config/deploy/context.xml.tmpl
+++ b/charts/pega/config/deploy/context.xml.tmpl
@@ -28,4 +28,7 @@
   <Environment name="prconfig/database/databases/CustomerData/defaultSchema" value="{{ .Env.DATA_SCHEMA }}" type="java.lang.String" />
   {{ end }}
   <Environment name="prconfig/database/drivers" value="com.microsoft.sqlserver.jdbc.SQLServerDriver;oracle.jdbc.OracleDriver;org.postgresql.Driver;com.ibm.db2.jcc.DB2Driver" type="java.lang.String" />
+  {{ if .Env.WEB_REQUESTOR_PASSIVATION_TIMEOUT }}
+  <Environment name="prconfig/timeout/browser" value="{{ .Env.WEB_REQUESTOR_PASSIVATION_TIMEOUT }}" type="java.lang.String" />
+  {{ end }}
 </Context>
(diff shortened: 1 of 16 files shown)


dcasavant:
Is any documentation change in the readme or community required for this? 

@taz-mon

nowam1:
> Is any documentation change in the readme or community required for this?
> 
> @taz-mon

Probably description of new option: requestorPassivationToAlbStickinessMargin is required

charts/pega/templates/_helpers.tpl
```diff
@@ -187,3 +187,14 @@ readinessProbe:
     Background
   {{- end -}}
 {{- end }}
+
+# Load balancer session cookie stickiness time in seconds,
+# calculated as sum of .requestor.passivationTimeSec and passivation delay.
+{{- define "lbSessionCookieStickiness" }}
```
dcasavant: Is it possible to add a check after the to see if `alb_stickiness_lb_cookie_duration_seconds` is set and use that value if the new configuration is not present?  That will preserve backwards compatibility and not require client changes of their values.yaml files.

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -28,4 +28,7 @@
   <Environment name="prconfig/database/databases/CustomerData/defaultSchema" value="{{ .Env.DATA_SCHEMA }}" type="java.lang.String" />
   {{ end }}
   <Environment name="prconfig/database/drivers" value="com.microsoft.sqlserver.jdbc.SQLServerDriver;oracle.jdbc.OracleDriver;org.postgresql.Driver;com.ibm.db2.jcc.DB2Driver" type="java.lang.String" />
+  {{ if .Env.WEB_REQUESTOR_PASSIVATION_TIMEOUT }}
+  <Environment name="prconfig/timeout/browser" value="{{ .Env.WEB_REQUESTOR_PASSIVATION_TIMEOUT }}" type="java.lang.String" />
```
dcasavant: Remove WEB_ and just call the variable REQUESTOR_PASSIVATION_TIMEOUT to match the Helm parameter?

nowam1: Changed the name to REQUESTOR_PASSIVATION_TIMEOUT

charts/pega/templates/_helpers.tpl
nowam1: Good idea. Added a check for alb_stickiness_lb_cookie_duration_seconds after checking that new configuration values are missing. If both new and old configs are present new will take precedence. Added one more check for presence of `requestor` block as it is missing in old configs and anyway it is optional.

Issue #27, opened 2019-10-22
Chart should be updated to support Kubernetes 1.16.x


rajshrimohanks:
Kubernetes v1.16.x has removed a lot of [deprecated API](https://kubernetes.io/docs/setup/release/notes/#deprecations-and-removals)s which this chart still seem to be using. They should be changed to their new versions in order to support Kubernetes v1.16.

dcasavant:
Agreed - added to the backlog.

dcasavant:
Merged.  This will be present in the next release pushed to bintray.

PR #33, opened 2019-11-25
Helm 3 / Kubernetes 1.16.x

dcasavant:
* Updates for Helm 3 #34 .
* Readme now references bintray repo and the *getting started* section no longer requires a repo clone.
* API references updated to support Kubernetes 1.16.x. #27 
* Fixed host name mapping for OpenShift's route
* Converted several sections to be parameterized
  * Added pegasearch.set_vm_max_map_count - Elasticsearch requires vm.max_map_count to be configured to an appropriate value.  This may be configured manually on your system and if so, you may bypass this privileged init container. For more information, see the [ElasticSearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html).  This value defaults to **true**.
  * Added pegasearch.runAsUser - Specify a user to run as.  This is helpful in locked down environments. Default runAsUser is **1000**.
  * Added pegasearch.set_data_owner_on_startup - Set to true to enable an init container that runs a chown command on the mapped volume at startup to reset the owner of the ES data to the current user.  This is needed if a random user is used to run the pod, but also requires privileges to change the ownership of files.  This value defaults to **true**.

### DIFF
--- a/charts/pega/templates/pega-tier-deployment.yaml
+++ b/charts/pega/templates/pega-tier-deployment.yaml
@@ -6,7 +6,7 @@
 {{ $apiVer = "apps/v1" }}
 {{ else }}
 {{ $kindName = "Deployment" }}
-{{ $apiVer = "extensions/v1beta1" }}
+{{ $apiVer = "apps/v1" }}
 {{ end }}
 
 {{ $containerWaitList := list  }}
--- a/charts/pega/templates/pega-tier-deployment.yaml
+++ b/charts/pega/templates/pega-tier-deployment.yaml
@@ -3,7 +3,7 @@
 {{ range $index, $dep := .Values.global.tier}}
 {{ if ($dep.volumeClaimTemplate) }}
 {{ $kindName = "StatefulSet" }}
-{{ $apiVer = "apps/v1beta2" }}
+{{ $apiVer = "apps/v1" }}
 {{ else }}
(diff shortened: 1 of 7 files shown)


charts/pega/templates/pega-tier-deployment.yaml
```diff
@@ -3,7 +3,7 @@
 {{ range $index, $dep := .Values.global.tier}}
 {{ if ($dep.volumeClaimTemplate) }}
 {{ $kindName = "StatefulSet" }}
-{{ $apiVer = "apps/v1beta2" }}
+{{ $apiVer = "apps/v1" }}
 {{ else }}
 {{ $kindName = "Deployment" }}
 {{ $apiVer = "extensions/v1beta1" }}
```
MadhuriArugula: Should we also modify the api for Deployment to use apps/v1 as per kubernetes documentation https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

rajshrimohanks: I tested this and it wouldn't work without this change. So I guess this should be changed.

```yaml
{{ $kindName = "Deployment" }}
{{ $apiVer = "apps/v1" }}
```

dcasavant: Fixed.  Good find!

PR #44, opened 2019-12-10
installer/jobs-reader role: limit rules

iMax-pp:


charts/pega/charts/installer/templates/pega-installer-role.yaml
```diff
@@ -5,7 +5,13 @@ metadata:
...
-- apiGroups: ["", "batch", "extensions", "apps"]
-  resources: ["jobs", "deployments", "statefulsets"]
+- apiGroups: ["apps"]
+  resources: ["deployments", "statefulsets"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["batch"]
+  resources: ["jobs"]
+  verbs: ["get", "watch", "list"]
+- apiGroups: ["extensions"]
+  resources: ["deployments"]
```
MadhuriArugula: @iMax-pp Is there any specific reason we are splitting the api groups and verbs?

charts/pega/charts/installer/templates/pega-installer-role.yaml
```diff
@@ -5,7 +5,13 @@ metadata:
   name: {{ template "installerJobReaderRole" }}
   namespace: {{ .Release.Namespace }}
 rules:
-- apiGroups: ["", "batch", "extensions", "apps"]
```
MadhuriArugula: I didn't the reason for removing core API group.

charts/pega/charts/installer/templates/pega-installer-role.yaml
iMax-pp: I splitted api groups and resources to allow for different verbs on each couple of api group / resource.

charts/pega/charts/installer/templates/pega-installer-role.yaml
iMax-pp: I restricted the rules, as it was not working on the cluster I'm deploying on. But with these restricted rights, the installer deployment is working fine.

PR #52, opened 2019-12-19
Check for null driverUri

### LINKED ISSUE #50: Error when omitting JDBC Driver URI
We're using an image with the driver baked in.  If we comment out jdbc.driverUri or give it a blank value we get the error:
DriverURIHelmError

dcasavant:
Check for a missing driverUri (not required) before generating. 

This fixes #50


dcasavant:
Confirmed to fix the reported issue.

dcasavant:
Accidentally closed without merging - fixing.

Issue #53, opened 2019-12-24
Service Session are not sticky when using NodePort


Emilio-Pega:
The service is not defined to be sticky when using a NodePort (as far as I can tell), this is causing unusual behavior in designer studio when I have replicas >= 2 on the web tie.  Manifests in a ll sorts of unusual ways:

1. sometimes fails to serve static content
2. Can't save-as rules
3. Cannot open some rule
4. Cannot execute some rules
5. Etc...

All these issues go away when I scale back to a single web node.

Based on this https://kubernetes.io/docs/concepts/services-networking/service/ I would like to see `service.spec.sessionAffinity` set to `ClientIP` *at least for the NodePort case*

This is important for producing a helm based SDE with multiple web nodes.

Pega bug: BUG-531671

Emilio-Pega:
After this happening for several days, could no longer reproduce the issue, so closing

Emilio-Pega:
Ok, it reproduced again for Piort, its somewhat inconsistent.  I am re-opening the issue.

On our local K8S I made the fix I proposed above and the issue cleared up

Emilio-Pega:
https://github.com/pegasystems/pega-helm-charts/pull/54

kishorv10:
@Emilio-Pega  Ingress with Traefik is the suggested way to achieve session affinity for Pega web nodes. Are you not using Traefik?

Emilio-Pega:
I have been told by @dcasavant to use `LoadBalancer` with Traefik 

Does this mean plain NodePort *without* a front-end load balancer is not supported by these charts?  

I know that k8s *will* load balance to a NodePort service without a front-end loadbalancer like Traefik.  So my expectation based on the readme's would be that this should/can work without something like Traefik.

If there is a section in the readme that I missed that says this is *not* supported can you please point it out, as I can't find it.

Emilio-Pega:
Its acceptable to me if you say this is not supported, provided that it is documented

kishorv10:
Per documentation, NodePort without a front-end load balancer (for pega-web service) is suggested only for minikube deployments. See [here](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/values-minimal.yaml#L69) 
[...]

PR #54, opened 2020-01-02
ISSUE-53 added session affinity flag to pega service

caleighlink1:


Emilio-Pega:
Proposed fix for https://github.com/pegasystems/pega-helm-charts/issues/53

charts/pega/templates/_pega-service.tpl
```diff
@@ -25,6 +25,7 @@ spec:
   - name: http
     port: {{ .port }}
     targetPort: {{ .targetPort }}
+  sessionAffinity: ClientIP
```
dcasavant: This would set sessionAffinity to ClientIP for everyone.  If you are using a load balancer, how does this work along side cookie based session affinity?

The default timeoutSeconds of ~3hrs would also need to be set based on the passivation timeout like is done for the ALB.

dcasavant:
Closing.  See comments on #53 .  Please reopen if needed.

PR #60, opened 2020-01-09
Configurable imagePullPolicy

### LINKED ISSUE #56: imagePullPolicy is not accurate/correct
The described `imagePullPolicy` in the readme appears to be inaccurate.  The templates (as far as I can tell) do not explicitly set `imagePullPolicy`

The readme:
Capture

The actual behavior if  `imagePullPolicy` is not specified https://kubernetes.io/docs/concepts/configuration/overview/#container-images

Basically, the readme is only accurate if you use a docker tag of `latest` or do not specify a docker tag.

dcasavant:
Add support for specifying an imagePullPolicy for the pega and search images.

This fixes #56

### DIFF
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -385,6 +385,7 @@ Use the `pegasearch` section to configure a deployment of ElasticSearch for sear
 Parameter   | Description   | Default value
 ---         | ---           | ---
 `image`   | Set the `pegasearch.image` location to a registry that can access the Pega search Docker image. The image is [available on DockerHub](https://hub.docker.com/r/pegasystems/search), and you may choose to mirror it in a private Docker repository. | `pegasystems/search:latest`
+`imagePullPolicy` | Optionally specify an imagePullPolicy for the search container. | `""`
 `replicas` | Specify the desired replica count. | `1`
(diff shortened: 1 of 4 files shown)


Emilio-Pega:
LGTM

charts/pega/README.md
```diff
@@ -124,6 +122,9 @@ docker:
     url: "YOUR_DOCKER_REGISTRY"
     username: "YOUR_DOCKER_REGISTRY_USERNAME"
     password: "YOUR_DOCKER_REGISTRY_PASSWORD"
+  pega:
+    image: "pegasystems/pega"
+    imagePullPolicy: "Always"
```
cdancy: Why do we want `Always` as the default? It will make an unnecessary attempt to download the same image that may already be present locally.

Emilio-Pega: I don't think this PR changes it to always by default, thats just the example.  The README incorrectly says its Always by default which is what I mentioned i nthe original issue

WE (SQuID) need this beecause we are constantly refreshing tha specific packed web image (not latest  tag) for the helm SDEs

Emilio-Pega: SO Dave C has kindly provided us access to this value for our use case

Emilio-Pega: II also suspect that if a customer is using a non-latest tag for their image, they will want to use this option to make sure the get the latest tagged version with all of the security updates (since we periodically refresh tagged versions with security updates)

dcasavant: Correct that this is just the readme.  The default is not having an `imagePullPolicy` set.

PR #69, opened 2020-01-20
Support for native lb for google cloud. Service part.

Max-Levitskiy:


charts/pega/templates/gke/affinity-session-backendconfig.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: BackendConfig
+metadata:
+  name: gke-affinity-session-backendconfig
+spec:
+  timeoutSec: 40
+  connectionDraining:
+    drainingTimeoutSec: 60
+  sessionAffinity:
+    affinityType: "GENERATED_COOKIE"
+    affinityCookieTtlSec: 600
```
Heshma: This value should not be hardcoded. We should be calculating cookie expiration based on requestor passivation time. We added changes related to this when we added support for alb ingress controller for aws. We can reuse the same value here.

Max-Levitskiy: Fixed

PR #70, opened 2020-01-20
Ingress changes for GKE

mieka1979:
Ingress changes for GKE.


charts/pega/templates/_pega-gke-ingress.tpl
```diff
@@ -0,0 +1,19 @@
+{{- define "pega.gke.ingress" -}}
+# Ingress to be used for {{ .name }}
+kind: Ingress
+apiVersion: extensions/v1beta1
+metadata:
+  name: {{ .name }}
+  namespace: {{ .root.Release.Namespace }}
+spec:
+  rules:
```
saikarthik528: Please add default backend here 

backend:
  serviceName: {{ .name }}
  servicePort: {{ .node.service.port }}


Default backend is needed for redirection purposes in case the url doesn't match with any of the resources available.
Reference : https://stackoverflow.com/questions/57868848/problems-configuring-ingress-with-cookie-affinity

mieka1979: Fixed

PR #78, opened 2020-01-22
Add support for native Azure and Google load balancers

### LINKED ISSUE #42: GKE Ingress
GKE ingress now support session based affinity, https://cloud.google.com/load-balancing/docs/backend-service#session_affinity

Is there a limitation with this and Pega still?

Also GKE has a managed certificate resource: https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs

So ideally we would just use Google to manage all that stuff for us and get rid of Traefik. Keep the deployment easier. If we can use ingress like I mention above let me know and I'll clone this repo and see if I can figure out the helm stuff.

Max-Levitskiy:
Closes #42


charts/pega/templates/_pega-eks-ingress.tpl
```diff
@@ -7,7 +7,7 @@ metadata:
   namespace: {{ .root.Release.Namespace }}
   annotations:
     # Ingress class used is 'alb'
-    kubernetes.io/ingress.class: alb
+    kubernetes.io/ingress.class: {{ include "ingressClass" . }}
```
kishorv10: Looks like, this template is specific to EKS, Can we hard code the ingress.class ```alb```, instead of deriving?

charts/pega/templates/_pega-k8s-ingress.tpl
```diff
@@ -13,7 +13,11 @@ metadata:
 {{- end }}
 {{- else }}
     # Ingress class used is 'traefik'
-    kubernetes.io/ingress.class: traefik
+    kubernetes.io/ingress.class: {{ include "ingressClass" . }}
```
kishorv10: I recommend creating AKS specific ingress tpl instead of conditioning in k8s-ingress tpl as we have provider specific ingress templates for GKE & EKS. To re-use the code blocks, consider generating below piece of code from helpers tpl. This seems to be common across all providers
```
 rules:
  # The calls will be redirected from {{ .node.domain }} to below mentioned backend serviceName and servicePort.
  # To access the below service, along with {{ .node.domain }}, http/https port also has to be provided in the URL.
  - host: {{ .node.service.domain }}
    http:
      paths:
      - backend:
          serviceName: {{ .name }}
          servicePort: {{ .node.service.port }}
```

saikarthik528: There are significant changes for other ingress definitions, but, for aks, this seems to be the only change.  Hence, we went with using the same file.  Better yet, I will create a new file and move it since it seems to be the common approach.

charts/addons/values.yaml
```diff
@@ -117,3 +118,30 @@ metrics-server:
 # Not recommended for production usage, but can be useful in test clusters with self-signed Kubelet serving certificates.
 # This setting is not required for public cloud providers & openshift enterprise. It may be required for open-source Kubernetes.
 # - --kubelet-insecure-tls
+
+# Set this to true to install application-gateway-ingress-controller. Follow below guidelines specific to each provider,
```
dcasavant: Th addons readme needs to be updated with the changes for GKE and AKS.

charts/pega/templates/_pega-aks-ingress.tpl
```diff
@@ -0,0 +1,24 @@
...
+kind: Ingress
+apiVersion: extensions/v1beta1
+metadata:
+  name: {{ .name }}
+  namespace: {{ .root.Release.Namespace }}
+  annotations:
+    # Ingress class used is 'azure/application-gateway'
+    kubernetes.io/ingress.class: azure/application-gateway
+    # Ingress annotations for aks
+    appgw.ingress.kubernetes.io/cookie-based-affinity: "true"
```
dcasavant: Is there a cookie timeout we can apply to this ingress?

docs/Deploying-Pega-on-GKE.md
```diff
@@ -370,7 +333,64 @@ $ kubectl create namespace pegaaddons
...
+          secretName:
+          useManagedCertificate: false
+          ssl_annotation:
+             ingress.gcp.kubernetes.io/pre-shared-cert: webCert
+```
+
+- To support HTTPS connectivity with Pega Platform, you configure a managed SSL certificate and associate it with the ingress by making the following changes in the pega.yaml file for the exposed tiers in your deployment:
+
+```yaml
+ingress:
```
saikarthik528: This needs to change.  For managed certificate, the following yaml stub should be used :

ingress:
          domain: "webtest.gketest1.dev.pega.io"
          tls:
            enabled: true
            secretName:
            useManagedCertificate: true
            ssl_annotation:
              # This is not mandatory.  We can create a static IP address and use it in any of the approaches using this annotation.
              kubernetes.io/ingress.global-static-ip-name: lb-ip-address

taz-mon: updated the step and incorporated your review comments

charts/addons/values.yaml
```diff
@@ -1,66 +1,74 @@
...
-  enabled: true
-  # Set any additional Traefik parameters. These values will be used by Traefik's helm chart.
+  enabled: false
+  # When you set this to true, you can then set additional Traefik parameters to be passed into Traefik's Helm chart.
   # See https://github.com/helm/charts/blob/master/stable/traefik/values.yaml
-  # Set traefik.serviceType to "LoadBalancer" on gke, aks, and pks
-  serviceType: NodePort
-  # If enabled is set to "true", ssl will be enabled for traefik
+  # To use Traefik as a load balancer in PKS, AKS, GKE, or k8s, set traefik.serviceType: "LoadBalancer".
+  serviceType: LoadBalancer
```
dcasavant: Should this be `NodePort` by default?  The comment above says to _change_ it to `LoadBalancer` iff using specified cloud providers.

charts/pega/values.yaml
```diff
@@ -76,11 +76,29 @@ global:
...
+        # https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/README.md#ingress
+
         # Enter the domain name to access web nodes via a load balancer.
         #  e.g. web.mypega.example.com
         domain: "YOUR_WEB_NODE_DOMAIN"
-        port: 80
-        targetPort: 8080
+        tls:
+          # Enable TLS encryption
+          enabled: true
```
dcasavant: Should this be `false` by default since it requires extra config?

charts/pega/values.yaml
```diff
@@ -76,11 +76,29 @@ global:
...
+          secretName:
```
dcasavant: Should this be omitted if not set?

charts/pega/values.yaml
```diff
@@ -76,11 +76,29 @@ global:
...
+          # For GKE Managed Certificate, mention true if Google Managed Certificate has to be created and annotation specified
+          useManagedCertificate: false
+          # Provide appropriate certificate annotations for EKS or GKE
+          # For EKS, use alb.ingress.kubernetes.io/certificate-arn: <certificate-arn>
+          # For GKE Pre-shared Certificate, use ingress.gcp.kubernetes.io/pre-shared-cert: <pre-shared-certificate-name>
+          # For GKE to use static IP for load balancer, use kubernetes.io/ingress.global-static-ip-name: <global-static-ip-name>
+          ssl_annotation:
```
dcasavant: Should this be omitted if not set?

dcasavant:
Travis is reporting the following error: `The job exceeded the maximum log length, and has been terminated.` Might be related to the verbose logging.

sunnygoel8:
@dcasavant , will it only support Azure and Google load balancers ? Or others as well such as AWS, NSX-T and F5 ??

dcasavant:
> @dcasavant , will it only support Azure and Google load balancers ? Or others as well such as AWS, NSX-T and F5 ??

@sunnygoel8 ALB is already supported.  The others you mentioned are not native for cloud providers and while they could be configured if they support the requirements of Pega, they are not _automatically configured_ by the charts.

charts/pega/templates/_pega-aks-ingress.tpl
dcasavant: I don't see an API for it, so I submitted an issue against the Azure gateway project.

https://github.com/Azure/application-gateway-kubernetes-ingress/issues/734

charts/pega/README.md
```diff
@@ -20,27 +20,32 @@ Example for a kubernetes environment:
...
+
+For additional, required installation parameters, see the [Installer section](#install).
 
 Value             | Action
 ---               | ---
 deploy            | Start the Pega containers using an existing Pega database installation.
 install           | Install Pega Platform into your database without deploying.
 install-deploy    | Install Pega Platform into your database and then deploy.
-upgrade           | Upgrade the Pega Platform installation in your database.
-upgrade-deploy    | Upgrade the Pega Platform installation in your database, and then deploy.
```
alexlaverty, line 36: hi was there a reason why upgrade support was removed for this container?

PR #80, opened 2020-01-22
Introduced tier name and kubernetes specific DSS settings 

Mandeep-pega:
Passed Tier of a node to prpc so that it can be used in admin studio. The NODE_TIER variable is set in setenv.sh file of docker. 
Passed two DSS settings to prpc so that,
1) Requestor pools can be disabled by default in pega, and 
2) Admin studio can display information without node ids

### DIFF
--- a/charts/pega/templates/_pega-deployment.tpl
+++ b/charts/pega/templates/_pega-deployment.tpl
@@ -89,8 +89,6 @@ spec:
         - name: REQUESTOR_PASSIVATION_TIMEOUT
           value: "{{ .node.requestor.passivationTimeSec }}"
 {{- end }}
-        - name: NODE_TIER
-          value: {{ .tierName }}
 {{- if .custom }}
 {{- if .custom.env }}
         # Additional custom env vars
@@ -98,6 +96,9 @@ spec:
 {{- end }}
 {{- end }}
 {{ include "pega.jvmconfig" (dict "node" .node) | indent 8 }}
+        # Tier of the Pega node
+        - name: NODE_TIER
+          value: {{ .tierName }}
         envFrom:
         - configMapRef:
             name: {{ template "pegaEnvironmentConfig" }}
--- a/charts/pega/templates/_pega-deployment.tpl
+++ b/charts/pega/templates/_pega-deployment.tpl
@@ -86,7 +86,7 @@ spec:
         - name: NODE_TYPE
           value: {{ .nodeType }}
         - name: NODE_TIER
-          value: {{ .name }}
+          value: {{ .tierName }}
         - name: NODE_SETTINGS
           value: Pega-UIEngine/cloud/isNodeAgnosticAdminStudio=true;Pega-IntegrationEngine/EnableRequestorPools=false;
 {{- if .node.requestor }}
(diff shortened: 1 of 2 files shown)


charts/pega/templates/_pega-deployment.tpl
```diff
@@ -85,6 +85,10 @@ spec:
         # Node type of the Pega nodes for {{ .name }}
         - name: NODE_TYPE
           value: {{ .nodeType }}
+        - name: NODE_TIER
+          value: {{ .tierName }}
+        - name: NODE_SETTINGS
+          value: Pega-UIEngine/cloud/isNodeAgnosticAdminStudio=true;Pega-IntegrationEngine/EnableRequestorPools=false;
```
dcasavant: For node settings that we want to always be present in the Docker image, we should probably just add them [here](https://github.com/pegasystems/docker-pega-web-ready/blob/master/tomcat-conf/Catalina/localhost/prweb.xml)

Emilio-Pega: If you are going to force in NODE_SETTINGS you need to allow customization, the SDE uses SDE specific NODE_SETTINGS

Emilio-Pega: Currently we are setting NODE_SETTINGS as a custom env var in the SDE as per

```
tier:
  - name: my-tier
    custom:
      env:
        - name: MY_ENV_NAME
          value: MY_ENV_VALUE
```

Emilio-Pega: Ok, i misspoke, we actually customize by passing in a custom JAVA_OPTS value, but the principal is the same, if you are going to do this we need to be able to override

Emilio-Pega: Example from an SDE web pod

Capture

Mandeep-pega:
I have changed the PR to remove NODE_SETTINGS environment variable. Will raise a different PR in docker image as suggested by @dcasavant

pega-embate:
The test that are failing should be fixed before this is merged:

TestPegaStandardTierDeployment 2020-01-23T11:49:21Z command.go:158: 
--- FAIL: TestPegaStandardTierDeployment (0.07s)
    deployment_utility.go:123: 
        	Error Trace:	deployment_utility.go:123
        	            				deployment_utility.go:185
        	            				deployment_utility.go:77
        	            				deployment_utility.go:40
        	            				deploy_test.go:29
        	Error:      	Not equal: 
        	            	expected: "NODE_TIER"
        	            	actual  : "REQUESTOR_PASSIVATION_TIMEOUT"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-NODE_TIER
        	            	+REQUESTOR_PASSIVATION_TIMEOUT
        	Test:       	TestPegaStandardTierDeployment
FAIL

Mandeep-pega:
Fixed test cases. Please merge

charts/pega/templates/_pega-deployment.tpl
dcasavant: This has been removed.  Closing thread.

PR #82, opened 2020-01-22
Add support for probe configuration Issue/#65

### LINKED ISSUE #65: Configurable liveness and readiness probes
Some of the pega pods can take a long time to fully start and we've found that the [300 second](https://github.com/pegasystems/pega-helm-charts/blob/7c7545cc2276ea086e25b8c2d7e115449a50d3ca/charts/pega/templates/_helpers.tpl#L145) `initialDelaySeconds` is not sufficient and causes pods to be restarted prematurely. 
 
As a user of the pega helm charts I want to be able to configure the liveness and readiness probes.

dcasavant:
### Liveness and readiness probes

[Probes are used by Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) as a way to determine if an application is healthy.  They may be configured for liveness to determine if a Pod has entered a broken state, or for readiness to determine if the application is available to be exposed.  Probes may be configured for each tier independently.  If not explicitly configured, default probes will be used.  The following parameters may be used as part of a `livenessProbe` or `readinessProbe` configuration.

Parameter           | Description    | Default value
---                 | ---            | ---
`initialDelaySeconds` | Number of seconds after the container has started before liveness or readiness probes are initiated. | `300`
`timeoutSeconds`      | Number of seconds after which the probe times out. | `20`
`periodSeconds`       | How often (in seconds) to perform the probe. | `30`
`successThreshold`    | Minimum consecutive successes for the probe to be considered successful after having failed. | `1`
`failureThreshold`    | When a Pod starts and the probe fails, Kubernetes will try *failureThreshold* times before giving up. | `3`

Example:

```yaml
tier:
  - name: my-tier
      livenessProbe:
        initialDelaySeconds: 60
        timeoutSeconds: 30
        failureThreshold: 5
      readinessProbe:
        initialDelaySeconds: 400
        failureThreshold: 30
```

This closes #65

### DIFF
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -243,6 +243,30 @@ Parameter       | Description    | Default value
 `initialHeap`   | This specifies the initial heap size of the JVM.  | `4096m`
 `maxHeap`       | This specifies the maximum heap size of the JVM.  | `7168m`
 
+### Liveness and readiness probes
+
+[Probes are used by Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) as a way to determine if an application is healthy.  They may be configured for liveness to determine if a Pod has entered a broken state, or for readiness to determine if the application is available to be exposed.  Probes may be configured for each tier independently.  If not explicitly configured, default probes will be used.  The following parameters may be used as part of a `livenessProbe` or `readinessProbe` configuration.
+
+Parameter           | Description    | Default value
+---                 | ---            | ---
+`initialDelaySeconds` | Number of seconds after the container has started before liveness or readiness probes are initiated. | `300`
+`timeoutSeconds`      | Number of seconds after which the probe times out. | `20`
+`periodSeconds`       | How often (in seconds) to perform the probe. | `10`
+`successThreshold`    | Minimum consecutive successes for the probe to be considered successful after having failed. | `1`
+`failureThreshold`    | When a Pod starts and the probe fails, Kubernetes will try *failureThreshold* times before giving up. | `3`
+
+```yaml
+tier:
+  - name: my-tier
+      livenessProbe:
+        initialDelaySeconds: 60
+        timeoutSeconds: 60
+        failureThreshold: 5
+      readinessProbe:
+        initialDelaySeconds: 400
+        failureThreshold: 30
+```
+
 ### Using a Kubernetes Horizontal Pod Autoscaler (HPA)
 
 You may configure an HPA to scale your tier on a specified metric.  Only tiers that do not use volume claims are scalable with an HPA. Set `hpa.enabled` to `true` in order to deploy an HPA for the tier. For more details, see the [Kubernetes HPA documentation](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/). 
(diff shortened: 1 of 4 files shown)


charts/pega/templates/_pega-deployment.tpl
```diff
@@ -140,7 +140,30 @@ spec:
 {{- end }}
         - name: {{ template "pegaVolumeCredentials" }}
           mountPath: "/opt/pega/secrets"
-{{ include "pega.health.probes" .root | indent 8 }}
+      # LivenessProbe: indicates whether the container is live, i.e. running.
```
MadhuriArugula: There seems to be an indentation issue. "livenessProbe" should be parallel to volumeMounts.
Indenting the newly added lines with another 2 spaces should solve the build failures.

brettcurtis:
I'm working a case with google and timeoutSeconds can not be greater than periodSeconds. I'm hitting this error when creating my own gke ingress.

```Error during sync: error running backend syncing routine: error ensuring health check: googleapi: Error 400: Invalid value for field 'resource.timeoutSec': '20'. TimeoutSec should be less than checkIntervalSec., invalid```

https://developers.google.com/resources/api-libraries/documentation/compute/v1/csharp/latest/classGoogle_1_1Apis_1_1Compute_1_1v1_1_1Data_1_1HttpHealthCheck.html#a027a3932f0681df5f198613701a83145

I'm thinking you'll probably want to change the defaults.

File: charts/pega/README.md [hunk 2]
```diff
@@ -243,6 +243,32 @@ Parameter       | Description    | Default value
...
+
+Parameter           | Description    | Default value
+---                 | ---            | ---
+`initialDelaySeconds` | Number of seconds after the container has started before liveness or readiness probes are initiated. | `300`
+`timeoutSeconds`      | Number of seconds after which the probe times out. | `20`
+`periodSeconds`       | How often (in seconds) to perform the probe. Some providers such as GCP require this value to be greater than the timeout value. | `30`
+`successThreshold`    | Minimum consecutive successes for the probe to be considered successful after having failed. | `1`
+`failureThreshold`    | When a Pod starts and the probe fails, Kubernetes will try *failureThreshold* times before giving up. | `3`
```
taz-mon: The number of attempts Kubernetes will try to restart a failing pod before it gives up and reports the Pod start up has failed.

charts/pega/README.md
```diff
@@ -243,6 +243,32 @@ Parameter       | Description    | Default value
...
 
+### Liveness and readiness probes
+
+[Probes are used by Kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) as a way to determine if an application is healthy.  They may be configured for liveness to determine if a Pod has entered a broken state, or for readiness to determine if the application is available to be exposed.  Probes may be configured for each tier independently.  If not explicitly configured, default probes will be used.  The following parameters may be used as part of a `livenessProbe` or `readinessProbe` configuration.
+
+Parameter           | Description    | Default value
+---                 | ---            | ---
+`initialDelaySeconds` | Number of seconds after the container has started before liveness or readiness probes are initiated. | `300`
+`timeoutSeconds`      | Number of seconds after which the probe times out. | `20`
+`periodSeconds`       | How often (in seconds) to perform the probe. Some providers such as GCP require this value to be greater than the timeout value. | `30`
```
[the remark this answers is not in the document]
brettcurtis: Thanks looks good! GCP is fine since GKE will let you apply the probe but the error is coming from compute engine trying to create the health checks for the load balancer.

charts/pega/README.md
dcasavant: Not exactly what this is, but I am rewording it to match your other feedback.  See https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes for more info.

PR #83, opened 2020-01-23
Add NodePort as service type fo GKE.

mieka1979:
Add NodePort as service type fo GKE.


Max-Levitskiy:
@mieka1979 you should resolve the conflict and fix tests :)

PR #110, opened 2020-03-11
Fixing annotation bugs related to ingress changes

dcasavant:
This PR is related to two recent issues both related to ingress annotations.


dcasavant:
> Probably it will be a good idea to cover this place in tests. I think if we wanna fix it quickly we should merge it now but create an issue as tech debt.

Agreed.  I'll try to work some scheduling magic and maybe we can incorporate this into some housekeeping/refactoring between releases.

PR #112, opened 2020-03-11
ISSUE-111: Define PEGA_APP_CONTEXT_ROOT to define application root co…

pega-embate:
…ntext for pega tomcat containers

### DIFF
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -238,6 +238,7 @@ Specify the `ingress` yaml block to expose a Pega tier to access from outside Ku
 Parameter | Description
 ---       | ---
 `domain`  | Specify a domain on your network in which you create an ingress to the load balancer.
+`appContextPath`  | Specify the application context path to be used to access applications on a specific tier.  If not specified, /prweb will used. 
 `tls.enabled` | Specify the use of HTTPS for ingress connectivity. If the `tls` block is omitted, TLS will not be enabled.
 `tls.secretName` | Specify the Kubernetes secret you created in which you store your SSL certificate for your deployment. For compatibility, see [provider support for SSL certificate injection](#provider-support-for-ssl-certificate-management).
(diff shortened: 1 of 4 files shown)


charts/pega/templates/_pega-deployment.tpl
```diff
@@ -160,7 +163,7 @@ spec:
         {{- $readinessProbe := .node.readinessProbe }}
         readinessProbe:
           httpGet:
-            path: "/prweb/PRRestService/monitor/pingService/ping"
+            path: "/{{ .root.Values.global.appContextRoot | default "prweb" }}/PRRestService/monitor/pingService/ping"
```
cdancy: Would it be better to set this in our `helpers` file so that we're not repeating things multiple times and instead we can just reference a single property? Just thinking more down the road when we might want to change things for some reason and might miss one of these hard-coded `prweb` values.

pega-embate: Good idea @cdancy I will rework this to define/use a template in helpers with the app context setting from values (if that is not what you were proposing let me know)

charts/pega/README.md
```diff
@@ -238,6 +238,7 @@ Specify the `ingress` yaml block to expose a Pega tier to access from outside Ku
 Parameter | Description
 ---       | ---
 `domain`  | Specify a domain on your network in which you create an ingress to the load balancer.
+`appContextPath`  | Specify the application context path to be used to access applications on a specific tier.  If not specified, /prweb will used.
```
misterdorito, line 241: I'm assuming that you don't need to specify the leading '/' ?

pega-embate: Updated description to remove leading slash.  It is not needed.

pega-embate:
This pull request is being replaced by https://github.com/pegasystems/pega-helm-charts/pull/167

PR #118, opened 2020-03-24
finalize AGIK authentication details and add details for downloading docker images

taz-mon:
Made various changes:
- For all runbooks: moved addons documentation to the platform-specific runbook. Formerly it was in the prepping your system runbooks.
- For AKS deployments: updated AKS runbook and the /pega-helm-charts/tree/master/charts/addons values.yaml with clear on authentication, added the missing parameter, usePrivateIP: true
- Updated /pega-helm-charts/tree/master/charts/addons/readme.md with AGIK details.
- Added details for downloading Pega-generated docker images to the two prepping runbooks and added details in the /pega-helm-charts/tree/master/readme.md file

PR #120, opened 2020-03-26
adding default value to usePrivateIP

Heshma:
usePrivateIP doesn't have a default value because of which we are facing issue in deployment, commented on the PR - https://github.com/Azure/application-gateway-kubernetes-ingress/pull/722, where they claim it as issue fixed.


dcasavant:
This is also fixed in #118 but has a different value set.  @taz-mon can you sync with @Heshma to determine the correct value for the example and doc?

PR #126, opened 2020-04-06
ISSUE-125: Add option to set CATALINA_OPTS, with default values

maracle6:
Updated helpers to set ADDITIONAL_CATALINA_OPTS based on additionalCatalinaOpts. Updated README.md with appropriate instructions.

Default value of CATALINA_OPTS is set to recommended values for GC and heap (metaspace, code cache size, etc).

I implemented this as CATALINA_OPTS instead of JAVA_OPTS since JAVA_OPTS applies to the stop command and forked processes like Kafka.  I also fixed inconsistencies with indentation in a couple of ```yaml blocks in the README.md that was breaking the preview in Macdown...other editors might not mind.

This change needs to be paired with a small update to setenv.sh in the docker image.

[setenv.sh.zip](https://github.com/pegasystems/pega-helm-charts/files/4441076/setenv.sh.zip)

This also addresses https://github.com/pegasystems/docker-pega-web-ready/issues/37

### DIFF
--- a/charts/pega/templates/_helpers.tpl
+++ b/charts/pega/templates/_helpers.tpl
@@ -122,6 +122,12 @@ until cqlsh -u {{ $cassandraUser | quote }} -p {{ $cassandraPassword | quote }} 
 # Additional JVM arguments
 - name: JAVA_OPTS
   value: "{{ .node.javaOpts }}"
+- name: ADDITIONAL_CATALINA_OPTS
+{{- if .node.additionalCatalinaOpts }}
+  value: "{{ .node.additionalCatalinaOpts }}"
+{{- else }}
+  value: "-XX:MetaspaceSize=512M -XX:MaxMetaspaceSize=1G -XX:ReservedCodeCacheSize=512M -XX:+DisableExplicitGC -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication -Djava.awt.headless=true -Xlog:gc*,gc+heap=debug,gc+humongous=debug:file=/usr/local/tomcat/logs/gc.log:tags,uptime,pid,level,time"
+{{- end }}
 # Initial JVM heap size, equivalent to -Xms
 - name: INITIAL_HEAP
 {{- if .node.initialHeap }}
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -103,7 +103,7 @@ It is standard practice to have separate schemas for your rules and data.  You m
  
  Example:
  
- ```yaml
+```yaml
 jdbc:
   ...
   rulesSchema: "rules"
@@ -119,7 +119,7 @@ When using a private registry that requires a username and password, specify the
 
 Example:
 
- ```yaml
+```yaml
 docker:
   registry:
     url: "YOUR_DOCKER_REGISTRY"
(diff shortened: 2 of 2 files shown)


charts/pega/templates/_helpers.tpl
```diff
@@ -122,6 +122,12 @@ until cqlsh -u {{ $cassandraUser | quote }} -p {{ $cassandraPassword | quote }}
 # Additional JVM arguments
 - name: JAVA_OPTS
   value: "{{ .node.javaOpts }}"
+- name: ADDITIONAL_CATALINA_OPTS
+{{- if .node.additionalCatalinaOpts }}
+  value: "{{ .node.additionalCatalinaOpts }}"
+{{- else }}
+  value: "-XX:MetaspaceSize=512M -XX:MaxMetaspaceSize=1G -XX:ReservedCodeCacheSize=512M -XX:+DisableExplicitGC -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication -Djava.awt.headless=true -Xlog:gc*,gc+heap=debug,gc+humongous=debug:file=/usr/local/tomcat/logs/gc.log:tags,uptime,pid,level,time"
```
pega-embate, line 129: Would it make more sense to have the default values baked into the pega-web-ready docker image to minimize the risk of dropping these recommended params?  Also, the hardcoded path for gc logs in these params makes me think the default params/values should be delivered elsewhere.

NekhilKotha:
> All JVM options can be passed by setting "**javaOpts**" in values.yaml.
> 
> We append the settings given here in the docker image.
> Please refer https://github.com/pegasystems/docker-pega-web-ready/blob/6472cb42acaa4b9bfd9f99e4a93db3389a520e1b/tomcat-bin/setenv.sh#L6

Apologies I missed the reason why you did not use the JAVA_OPTS.

maracle6:
That makes sense.  Attached an updated setenv.sh (needs to be renamed). 

If it looks ok I can update _helpers.tpl as well.

[setenvv2.sh.zip](https://github.com/pegasystems/pega-helm-charts/files/4446626/setenvv2.sh.zip)

pega-embate:
That seems like a more appropriate place for these defaults, but out of curiosity, are these args appropriate for all environments?  Where does this list of "recommended" args come from?

maracle6:
We apply these across the board at all clients.  I can't think of any cases where we wouldn't use them, they should be appropriate for very large heaps as well.  Agop Touroyan has curated these over years of testing.

I'm sure there are some edge cases.  Would it be better to make them defaults in the image and also provide an optional helm parameter to suppress the defaults if there's a worry that this will cause the image to need customization?

pega-embate:
Are these used on PegaCloud envs as well?

charts/pega/templates/_helpers.tpl
NekhilKotha: Instead of defaulting these values in templates, these need to be passed by users by setting them in values.yaml because this is a web-ready image which acts as base image for fullybaked images across all releases.

We can provide the ability to pass the "additionalCatalinaOpts" and append it to CATALINA_OPTS in docker-pega-web-ready (https://github.com/pegasystems/docker-pega-web-ready/blob/master/tomcat-bin/setenv.sh)

MadhuriArugula:
@maracle6 Changes related to CATALINA_OPTS were fixed as part of #229 Could you please review and let us know if we are good to close this PR?

dcasavant:
Closing per comment from @MadhuriArugula . This was added in #126.  Please reopen if necessary.

PR #147, opened 2020-05-29
US-350314: allow for passing in a secrets path.

cdancy:


charts/pega/templates/pega-registry-secret.yaml
```diff
@@ -1,3 +1,19 @@
+{{- if .Values.global.docker.config.path }}
+kind: ExternalSecret
+apiVersion: kubernetes-client.io/v1
+metadata:
+  name: {{ template "pegaRegistrySecret" }}
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-install
+    "helm.sh/hook-weight": "5"
```
cdancy, line 9: Copied this, and really the entire solution, from our internal "datadog SCE".

charts/pega/templates/pega-registry-secret.yaml
```diff
@@ -1,3 +1,19 @@
...
+spec:
+  type: kubernetes.io/dockerconfigjson
+  backendType: secretsManager
+  data:
+    - key: {{ .Values.global.docker.config.path }}
```
cdancy, line 14: The cassandra chart uses `pullSecrets` as the name for a similar thing. We could adopt a similar naming pattern here as well and do something like `docker.config.pullSecrets`. Just throwing it out for discussion related purposes as I'm not married to the name here.

charts/pega/values.yaml
```diff
@@ -49,7 +49,10 @@ global:
       url: "YOUR_DOCKER_REGISTRY"
       username: "YOUR_DOCKER_REGISTRY_USERNAME"
       password: "YOUR_DOCKER_REGISTRY_PASSWORD"
-      # Docker image information for the Pega docker image, containing the application server.
+    # Optional config path to a secrets file if not using registry creds.
+    config:
+      path: ""
```
pega-embate, line 54: Can this param just be omitted or is the empty string significant?

pega-embate: If we leave the default empty string, can we add some comments about format and expected value of the path?  Can any path be specified or does the location need to be based on some existing root path?

cdancy: On the empty string: seen similar things done throughout this file and so copied that standard here for better/worse.

The "external-secrets" is an external project that can be found [HERE](https://github.com/godaddy/kubernetes-external-secrets) that we wrap internally as an SCE (ping me for name or just search "external" in bitbucket)

cdancy: I can put a comment to the relevant github project here and note how it's necessary for use if you're using this feature.

cdancy: @pega-embate just updated the documentation here to be as specific as possible. As this does require the use of an `kubernetes-externel-secrets`, which we use internally, I can see this _maybe_ being a potentially not desired change. If that's the case we could most likely modify our internal `SCE` to do this very thing if necessary instead of making the change here. I just figured this could be beneficial to the community.

@dcasavant any thoughts one way or another?

cdancy:
Closing this PR as we're taking this work internally so as not to confuse potential users/customers with this change and its requirements.

PR #167, opened 2020-07-17
ISSUE-111: Allow specification of application context path for each tier

pega-embate:


charts/pega/templates/_helpers.tpl
```diff
@@ -164,6 +164,19 @@ until cqlsh -u {{ $cassandraUser | quote }} -p {{ $cassandraPassword | quote }}
...
+{{- define "pega.applicationContextPath" -}}
+   {{- if .node.ingress -}}
+      {{- if .node.ingress.appContextPath -}}
+         {{ .node.ingress.appContextPath }}
+      {{- else -}}
+         prweb
+      {{- end -}}	 
+   {{- else -}}
+      prweb
+   {{- end -}}
```
pega-roska, line 177: Would it be possible to use an `and` here, or would that fail if there's no `.node.ingress`?
I'm thinking possibly something like this.
```suggestion
   {{- if and .node.ingress .node.ingress.appContextPath -}}
      {{ .node.ingress.appContextPath }}
   {{- else -}}
      prweb
   {{- end -}}
```

pega-embate: @pega-roska I agree that this is more concise, but helm seems to be evaluating .node.ingress.appContextPath regardless of the evaluation of .node.ingress to false.  I think I will leave this as-is.

dcasavant: Yea, Helm seems to be strange in that way.  I've banged my head against it in the past.  The only other workaround I have found to use is to set a local variable which will default to an object rather than _null_ so you can query it without failure.  See [the health probes](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/templates/_pega-deployment.tpl#L156) for an example.

terratest/src/test/pega/deployment_utility.go
```diff
@@ -118,6 +118,9 @@ func VerifyDeployment(t *testing.T, pod *k8score.PodSpec, expectedSpec pegaDeplo
 	var envIndex int32 = 0
 	require.Equal(t, pod.Containers[0].Env[envIndex].Name, "NODE_TYPE")
 	require.Equal(t, expectedSpec.nodeType, pod.Containers[0].Env[envIndex].Value)
+	envIndex++
+	require.Equal(t, pod.Containers[0].Env[envIndex].Name, "PEGA_APP_CONTEXT_PATH")
+	require.Equal(t, pod.Containers[0].Env[envIndex].Value, "prweb")
```
khick77, line 123: Does it make sense to add tests for values other than prweb?

pega-embate: Sure.  I will add a test for another value.

pega-embate:
> don't use passive voice...use this:
> Specify a context path applications use to access on a specific tier. If not specified, Pega Platform uses prweb.

@taz-mon 
How about this?
"Specify the path for access to the Pega application on a specific tier.  If not specified, users will have access to the Pega application using /prweb"

Issue #135, opened 2020-04-29
Unable to expose Kafka via default chart configuration


dcasavant:
**Describe the bug**
The default ingress configuration [hard codes the protocol as HTTP](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/templates/_pega-gke-ingress.tpl#L39).  This was found in GCP due to the ingress health check failing, but seems to apply to other providers as well.

**To Reproduce**
Deploy using the default configuration and view the stream health check.  Try to access the stream port externally.

**Expected behavior**
All Kubernetes objects are healthy.  Stream is accessible when configured.

**Chart version**
Latest

**Server (if applicable, please complete the following information):**
 - Environment: GCP

**Additional context**
In _addition_ to fixing the ability to deploy the stream tier, it may be easiest to disable its exposure by default in the [values.yaml](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/values.yaml#L149) file (and make any required runbook changes).

kishorv10:
Reopening the issue to fix below this [comment](https://github.com/pegasystems/pega-helm-charts/pull/173#pullrequestreview-463391802)

PR #173, opened 2020-08-05
ISSUE-135: Overriding health checks for web and stream load balancers

kishorv10:
This PR is to address [ISSUE-135](https://github.com/pegasystems/pega-helm-charts/issues/135)
In GKE, overriding health check of load balancer (specifically Stream) to use Platform ping service. This fix require GKE version to be 1.17.6-gke.11 or higher. Ref. [ingress-features](https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features).
This change does not introduce regression if deployed on lower versions than 1.17.6-gke.11.

### DIFF
--- a/charts/pega/templates/pega-gke-backend-config.yaml
+++ b/charts/pega/templates/pega-gke-backend-config.yaml
@@ -10,4 +10,12 @@ spec:
   sessionAffinity:
     affinityType: "GENERATED_COOKIE"
     affinityCookieTtlSec: {{ template "lbSessionCookieStickiness" }}
+  healthCheck:
+    checkIntervalSec: 5
+    healthyThreshold: 1
+    port: 8080
+    requestPath: /prweb/PRRestService/monitor/pingService/ping
+    timeoutSec: 5
+    type: HTTP
+    unhealthyThreshold: 2
 {{ end }}


misterdorito:
The current FCP version depends on this web app context rewrite feature.

kishorv10:
I will re-open the issue to fix the hardcoded "prweb" in the health url.

kishorv10:
> The application context root should not be hardcoded to "prweb". This healthcheck will fail if a custom app context is specified for a deployment.
> 
> pega.applicationContextPath defined in _helpers.tpl should be used instead:
> 
> e.g.
> requestPath: "/{{ template "pega.applicationContextPath" . }}/PRRestService/monitor/pingService/ping"
> 
> Although this will work ootb with the default configuration, this should be updated to avoid issues in the future.

{{ template "pega.applicationContextPath" . }} is tier specific however GKE backend config is one per the cluster. Do we have use cases to have separate app context path for each tier. Is it not same across all nodes of cluster? In that case, can this parameter be moved to global level instead tier (or node) specific?

pega-embate:
@kishorv10  We had an explicit request to support the application context path on a per tier basis to support different ingresses for each tier.  @misterdorito Can you provide additional information regarding this requirement?

misterdorito:
Currently web and stream tiers have ingresses/externally exposed endpoints.  Additionally I remember in the PR for the web app context overrides, that the prospect of adding additional ingresses for diagnostic purposes (for batch, etc.) -- I believe this was brought up in conjunction with SDE environments.

kishorv10:
Ok, this implies separate backend configs need to be generated for each ingress in GKE. I shall work on it.

PR #178, opened 2020-08-28
ISSUE-177 Add helm hook to force helm to wait until the install is complete

### LINKED ISSUE #177: Install and Upgrade helm commands do not wait until completion
**Describe the bug**
When using helm to execute the "install" or "upgrade" commands, the helm process terminates immediately on having deployed all resources. This makes it more difficult to use the helm chart within other pipelines and also makes it harder to tell when the install has completed. Using "install-deploy" and "upgrade-deploy" do wait for the install or upgrade process to complete, so this behavior seems inconsistent.

**To Reproduce**
Run the helm charts with `actions.execute` set to "install" or "upgrade".

**Chart version**
1.2.8

pega-talba:
Fixes #177 

Previously, running the install or upgrade action would exit helm as soon as all resources were deployed, not when the job was completed. Adding these hooks causes helm to wait until the resources are fully deployed. Note that we can only include install and upgrade (and not any deploy-related action) because otherwise we have a cyclical dependency (post-install hook + waiting for install to complete).

### DIFF
--- a/charts/pega/charts/installer/templates/_pega-installer-job.tpl
+++ b/charts/pega/charts/installer/templates/_pega-installer-job.tpl
@@ -5,6 +5,13 @@ apiVersion: batch/v1
 metadata:
   name: {{ .name }}
   namespace: {{ .root.Release.Namespace }}
+{{- if or (eq .root.Values.global.actions.execute "install") (eq .root.Values.global.actions.execute "upgrade") }}
+  annotations:
+    # Forces Helm to wait for the install or upgrade to complete.
+    "helm.sh/hook": post-install
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation
+{{- end }}
 spec:
   backoffLimit: 0
   template:
--- a/charts/pega/charts/installer/templates/_pega-installer-job.tpl
+++ b/charts/pega/charts/installer/templates/_pega-installer-job.tpl
@@ -5,7 +5,7 @@ apiVersion: batch/v1
 metadata:
   name: {{ .name }}
   namespace: {{ .root.Release.Namespace }}
-{{- if or (eq .root.Values.global.actions.execute "install") (eq .root.Values.global.actions.execute "upgrade") }}
+{{- if and .root.Values.waitForJobCompletion (or (eq .root.Values.global.actions.execute "install") (eq .root.Values.global.actions.execute "upgrade")) }}
   annotations:
     # Forces Helm to wait for the install or upgrade to complete.
     "helm.sh/hook": post-install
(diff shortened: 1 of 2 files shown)


dcasavant:
Can this be behind an optional flag?  That way people can kick and install synchronously or asynchronously.

pega-talba:
> Can this be behind an optional flag? That way people can kick and install synchronously or asynchronously.

Personally I would rather it not be optional - the previous behavior can make it appear that the process was successful even if the install or upgrade fails, and we are already adding many extra options when one of the hopes of the helm chart is to reduce and simplify things.

But, if we believe people may be relying on the current behavior, we can add an extra setting on the values yaml.

dcasavant:
@pega-talba I'm actually thinking more about how long it will take to install.  You don't get the logs piped back to your local machine, so it would potentially look like the process has hung for an extended period of time.  What happens if someone kills the local process while it is waiting? I'm guessing that would not impact the install?

pega-talba:
> 
> 
> @pega-talba I'm actually thinking more about how long it will take to install. You don't get the logs piped back to your local machine, so it would potentially look like the process has hung for an extended period of time. What happens if someone kills the local process while it is waiting? I'm guessing that would not impact the install?

This is a fair point, and given that there are already many people using this, we decided to add the setting after all. Updated the PR with a flag on the installer chart, which allows you to opt-in to waiting for the installer job to complete.

dcasavant (approved):

charts/pega/charts/installer/templates/_pega-installer-job.tpl
```diff
@@ -5,6 +5,13 @@ apiVersion: batch/v1
 metadata:
   name: {{ .name }}
   namespace: {{ .root.Release.Namespace }}
+{{- if and .root.Values.waitForJobCompletion (or (eq .root.Values.global.actions.execute "install") (eq .root.Values.global.actions.execute "upgrade")) }}
+  annotations:
+    # Forces Helm to wait for the install or upgrade to complete.
+    "helm.sh/hook": post-install
+    "helm.sh/hook-weight": "0"
+    "helm.sh/hook-delete-policy": before-hook-creation
```
pega-embate, line 13: How was the delete-policy chosen?  Would hook-succeeded make more sense?  (I'm not sure I fully understand how this policy is used)

pega-talba: There are three hook policies.

From https://helm.sh/docs/topics/charts_hooks/:
Annotation Value | Description
-- | --
before-hook-creation | Delete the previous resource before a new hook is launched (default)
hook-succeeded | Delete the resource after the hook is successfully executed
hook-failed | Delete the resource if the hook failed during execution

Originally we had hook-succeeded, but then it was brought up that unless the user has a log manager to push install logs away from Kubernetes, these would be lost immediately when the install finishes. We figure it is safer to keep the job around so people can fetch those logs as they please, as we often want to keep them even on success (especially upgrade logs). This also more closely emulates existing behavior.

PR #188, opened 2020-09-15
Add support from specifying additional volumes, volume mounts and env…

pegadave:
Add support from specifying additional volumes, volume mounts and environment vars to to the installer job in an effort to integrate with external secrets.

This PR (in conjunction with changes to the installer-ready-image) allows you to source the secrets used by the pega installer from a 3rd party secret management tool when subcharting the pega-helm-charts.  In order to take advantage of this:
--Include pega-helm-chart as a subchart.
--Override the "customInstallerVolumes" template in your chart to provide custom volumes sourced from your secret value:
```
{{- define "customInstallerVolumes" }}
- name: my-external-secret-volume1
  secret:
    secretName: my-external-secret1
    defaultMode: 420
{{- end }}
```
--Override the "customInstallerVolumeMounts" template in your chart to mount your volume:
```
{{- define "customInstallerVolumeMounts" }}
- name: my-external-secret-volume1
  mountPath: /opt/pega/ext-secret1
{{- end }}
```
--Override the "customInstallerEnvEntries" template in your chart so that you specify the mountPath of your volume:
```
{{- define "customInstallerEnvEntries" }}
- name: DB_USERNAME_SECRET_PATH
  value: /opt/pega/ext-secret1
{{- end }}
```

This PR allows you to specify both DB_USERNAME_SECRET_PATH or DB_PASSWORD_SECRET_PATH only.

### DIFF
--- a/charts/pega/charts/installer/templates/_helpers.tpl
+++ b/charts/pega/charts/installer/templates/_helpers.tpl
@@ -92,24 +92,4 @@
 {{- end }}
 {{- end }}
 
-#Override this template to define custom volumes (do not indent)
-{{- define "customInstallerVolumes" }}
-{{- end }}
-
-#Override this template to define custom volume mounts (do not indent)
-{{- define "customInstallerVolumeMounts" }}
-{{- end }}
-
-#Override this template to define environment entries (do not indent)
-{{- define "customInstallerEnvEntries" }}
-{{- end }}
-
-{{- define "customInstallerEnv" }}
-{{- $customEnvEntries := include "customInstallerEnvEntries" . }}
-{{- if ne "" $customEnvEntries }}
-env:
-{{- include "customInstallerEnvEntries" . }} 
-{{- end }}
-{{- end }}
-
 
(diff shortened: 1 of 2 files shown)


kishorv10 (commented):
@pegadave What is the reason behind providing overriding templates instead of exposing parameters in values.yaml. Supplying additional volumes, volumes mounts or environment variables can be exposed as new optional parameters in the values.yaml and render them in the templates. In this way, one need to modify the code to set them.

misterdorito:
@kishorv10 In my particular scenario, I am already subcharting the pega helm charts (so I am already writing code), but I do agree with you that would be more generically useful.  I will update the PR accordingly.

misterdorito:
So I have amended the PR so that specifying the following config in the values.yaml file will produce a similar result:
```
installer:
  image: "YOUR_INSTALLER_IMAGE:TAG"
  adminPassword: "ADMIN_PASSWORD"     

  custom:
    env:
    - name: "DB_PASSWORD_SECRET_PATH"
      value: "/opt/pega/ext-secret1"
    volumes:
    - name: my-secret-volume1
      secret:
        secretName: my-external-secret1
        defaultMode: 420
    volumeMounts:
    - name: "my-external-secret1"
      mountPath: "/opt/pega/ext-secret1"
```

This way the changes will be readily available even when you're not subcharting the pega helm charts.

misterdorito:
Abandoning this approach as we have come to an alternate approach which is more straightforward and doesn't involve changing each of the docker images.

PR #199, opened 2020-10-19
Ports 7003 exposed only on large installs

ragsns:
Ports 7003 exposed only on large installs and turned off by default on all others.


ragsns:
Fixed comments

kishorv10:
Hi @ragsns Port 7003 is not turned off by default. It is hidden from the default configuration. Chart deployment with default values.yaml, creates stream service at 7003 port. Are you not seeing this behavior?
User can set or pass these values during helm deployment to overwrite.

ragsns:
> Hi @ragsns Port 7003 is not turned off by default. It is hidden from the default configuration. Chart deployment with default values.yaml, creates stream service at 7003 port. Are you not seeing this behavior?

The suggestion is to not create stream service port at 7003 for all installs except the large installs, which is why it's commented. It can be left out with comments if required.

> User can set or pass these values during helm deployment to overwrite.

Ofcourse, but the default values should be reflective of most common install use cases, right?

dcasavant (approved):

charts/pega/values-large.yaml
```diff
@@ -148,6 +148,8 @@ global:
         # Inactivity time after which requestor is passivated
         passivationTimeSec: 900
 
+      # Port 7003 turned ON by default in all installs except large.
+      # comment following lines NOT to expose service and ports.
```
taz-pega-work, line 152: Port 7003 is exposed by default in all installs except large deployments.
 To NOT to expose service and ports in your deployment, you must comment out the following lines.

charts/pega/values.yaml
```diff
@@ -149,9 +149,11 @@ global:
         # Inactivity time after which requestor is passivated
         passivationTimeSec: 900
 
-      service:
-        port: 7003
-        targetPort: 7003
+      # Port 7003 turned off by default in all installs except large.
+      # Uncomment following lines to expose service and ports.
```
taz-pega-work, line 153: Port 7003 is not exposed by default in most deployments; the exception is large deployments.
To expose service and ports in your deployment, you must uncomment the following lines.

taz-pega-work (changes requested):
We should update the wording a bit to ensure that clients understand the defaults and how to over-write them

charts/pega/values.yaml
Saurabh-16: What is the reason we want to keep port and target port for stream commented in main values.yaml and only to have them by default in large values.yaml

ragsns:
In conversations with some customers, this is what we've found.

Most clients don’t use direct stream injection which is the reason for that ingress. The only real use case is in a CDH deployment.

charts/pega/values.yaml
ragsns: In conversations with some customers, this is what we've found.

Most clients don’t use direct stream injection which is the reason for that ingress. The only real use case is in a CDH deployment (primarily larger deployments). Rather than make it available across all deploys, we make it available for the most applicable cases and be able to turn off and on fairly easily by customer if they want to.

Saurabh-16: Please change the comment description as suggested by Taz.

ragsns:
Updated comments for Ports 7003 exposed only on large installs

ragsns:
Updated comments based on Taz's suggestion

MadhuriArugula:
@ragsns Can we close this PR as it is duplicate of #236

kishorv10:
Since there has been no activity on this pull request for the past year, I will proceed to close it. Please feel free to reopen it if necessary.

PR #203, opened 2020-10-29
Search service integration

pega-roska:
Add support for configuring the chart to use the external search service.

### DIFF
--- a/charts/pega/templates/pega-environment-config.yaml
+++ b/charts/pega/templates/pega-environment-config.yaml
@@ -47,6 +47,6 @@ data:
   CASSANDRA_NODES: "{{ include "cassandraNodes" . }}"
   # Port to connect to cassandra with
   CASSANDRA_PORT: "{{ .Values.dds.port }}"
-  # Deployment ID for FNX services
+  # Unique ID for multitenant backing services
   CUSTOMER_DEPLOYMENT_ID: "{{ include "customerDeploymentID" . }}"
 {{ end }}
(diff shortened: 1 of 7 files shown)


charts/pega/templates/pega-environment-config.yaml
```diff
@@ -32,12 +32,21 @@ data:
...
   # URL to connect to Elastic Search
   PEGA_SEARCH_URL: {{ .Values.pegasearch.externalURL }}
+{{ end }}
   # Whether to enable connecting to a cassandra cluster.  "true" for enabled, "false for disabled"
   CASSANDRA_CLUSTER: "{{ include "cassandraEnabled" . }}"
   # Comma separated list of cassandra hosts
   CASSANDRA_NODES: "{{ include "cassandraNodes" . }}"
   # Port to connect to cassandra with
   CASSANDRA_PORT: "{{ .Values.dds.port }}"
+  # Deployment ID for FNX services
```
misterdorito, line 50: Does this belong here?

pega-roska: Replaced with more generic phrasing

taz-pega-work (commented):
I believe you can clarify this cluster of statements to be active voice and group action:
# Only attempt to generate this file IF the user has NOT supplied
# an external search URL or enabled the search service, which means we
# need to deploy Elasticsearch internally.
===>
The deployment generates this file to invoke Elasticsearch internally when the external search URL is blank (or the default value is left) AND no search service is enabled.

pega-roska:
> I believe you can clarify this cluster of commented statements to be active voice and group the context clearly:
> Only attempt to generate this file IF the user has NOT supplied
> an external search URL or enabled the search service, which means we
> need to deploy Elasticsearch internally.
> ===>
> The deployment generates this file to invoke Elasticsearch internally when the external search URL is blank (or the default value is left) AND no search service is enabled.

This should be addressed, now.

misterdorito (approved):

pega-talba (approved):

PR #210, opened 2020-11-17
Feature/us 389604 - hazelcast client server architecture

Saurabh-16:
Adding new hazelcast subchart to ensure infinity can talk to hazelcast using client server model.


charts/pega/charts/hazelcast/templates/_helpers.tpl
```diff
@@ -0,0 +1,38 @@
...
+- name: wait-for-hazelcast
+  image: busybox:1.31.0
+  # Init container for waiting for hazelcast service to initialize.
+  args:
+  - sh
+  - -c
+  - >
+    set -e
+    counter=0;
+    while [ $(wget -q -O - "http://hazelcast-service.{{ .Release.Namespace }}:5701/hazelcast/health/cluster-size" /dev/null) -ne {{ .Values.hazelcast.replicas }} ] || [ $(wget -S "http://hazelcast-service.{{ .Release.Namespace }}:5701/hazelcast/health/cluster-safe" 2>&1 | grep "HTTP/" | awk '{print $2}') -ne 200 ]; do
```
kishorv10, line 24: Service name can be taken from template like done at other places.  ` {{ template "hazelcastName" }}-service `

charts/pega/charts/hazelcast/templates/hz-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "hazelcastName" . }}
+    component: Pega
+spec:
+  ports:
+    - port: 5701
+      targetPort: 5701
+      name: hazelcast
+  clusterIP: None
```
kishorv10, line 15: Is this deliberately made ` None `

charts/pega/charts/hazelcast/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+---
+apiVersion: v1
+description: Helm chart for Pega installation on kubernetes
```
kishorv10, line 3: Description can add details about hazlecast.

charts/pega/templates/pega-environment-config.yaml
```diff
@@ -54,5 +54,16 @@ data:
   PEGA_STREAM_URL: {{ .Values.stream.url }}
 {{- end }}
 {{- end }}
-
+{{ if .Values.hazelcast.enabled }}
+  ON_PEGA_K8s: "true"
```
kishorv10, line 58: Here ` s ` you may want in CAPS.

kishorv10 (approved):
Few general comments.

kishorv10 (commented):
Added few general comments. It would be good to add helm unit test cases for the hazlecast subchart.

charts/pega/templates/pega-environment-config.yaml
Saurabh-16: Addressed.

charts/pega/charts/hazelcast/Chart.yaml
Saurabh-16: Done.

charts/pega/charts/hazelcast/templates/hz-service.yaml
Saurabh-16: Yes. This service is headless service.

charts/pega/charts/hazelcast/templates/_helpers.tpl
Saurabh-16: I will check this

Saurabh-16: Done.

charts/pega/charts/hazelcast/templates/_helpers.tpl
```diff
@@ -0,0 +1,36 @@
...
+  - -c
+  - >
+    set -e
+    counter=0;
+    while [ -z $(wget -S "http://{{ template "hazelcastName" . }}-service.{{ .Release.Namespace }}:5701/hazelcast/health/cluster-safe" 2>&1 | grep "HTTP/" | awk '{print $2}') ] || [ $(wget -q -O - "http://{{ template "hazelcastName" . }}-service.{{ .Release.Namespace }}:5701/hazelcast/health/cluster-size" /dev/null) -ne {{ .Values.hazelcast.replicas }} ] || [ $(wget -S "http://{{ template "hazelcastName" . }}-service.{{ .Release.Namespace }}:5701/hazelcast/health/cluster-safe" 2>&1 | grep "HTTP/" | awk '{print $2}') -ne 200 ]; do
+    echo "waiting for hazelcast pods to start and join the cluster..." ;
+    counter=$(($counter+5));
+    sleep 5;
+    if [ $counter -gt 150 ]; then
+    echo "Timeout Reached. Hazelcast pods failed to join the cluster";
```
pega-embate, line 29: It might be worth clarifying this message.  To pass this test, we just need at least one hz server node up and running, right?  Maybe change this to: "Timeout Reached. Hazelcast pods are not available"

Also doesn't cluster-safe check to see if all data migrations are complete.  Is this required before allowing clients to connect to the hz cluster?

charts/pega/templates/_helpers.tpl
```diff
@@ -227,6 +227,9 @@ true
 {{- define "generatedPodAnnotations" }}
 {{- end }}
 
+{{- define "generatedHazelcastServicePodAnnotations" }}
```
pega-embate, line 230: Why is this here and not in the hazelcast _helpers.tpl?

Also, are datadog annotations being added later?  They should be defined here, right?

charts/pega/templates/pega-environment-config.yaml
```diff
@@ -54,5 +54,14 @@ data:
...
-
+{{ if .Values.hazelcast.enabled }}
+  HZ_DISCOVERY_K8S: "true"
+  HZ_CLIENT_MODE: "true"
+{{ if .Values.hazelcast.client.clusterName }}
+  # Hazelcast Cluster Name
+  HZ_CLUSTER_NAME: {{ .Values.hazelcast.client.clusterName }}
+{{- end }}
+  # DNS of hazelcast service
+  HZ_DNS: {{ template "hazelcastName" . }}-service.{{ .Release.Namespace }}.svc.cluster.local
```
pega-embate, line 65: DNS is not clear in this context.  Maybe HZ_SERVER_HOSTNAME would be better.

charts/pega/values.yaml
```diff
@@ -219,3 +219,11 @@ installer:
     upgradeType: "in-place"
     # Specify target rules schema for migration and upgrade
     targetRulesSchema: ""
+
+# Hazelcast settings
```
pega-embate, line 223: Are these settings needed both here and in the hazelcast values.yaml?

charts/pega/charts/hazelcast/values.yaml
```diff
@@ -0,0 +1,29 @@
...
+    cpu: 2
+  limits:
+    memory: "4Gi"
+    cpu: 2
+# client settings for hazelcast
+client:
+  clusterName: "PRPC"
+# Server side settings for hazelcast
+server:
+  java_opts: "-Dhazelcast.initial.min.cluster.size=3 -Djava.util.logging.config.file=/opt/hazelcast/logging.properties -Dhazelcast.jmx=true -XX:MaxRAMPercentage=80.0"
```
pega-embate, line 21: Have you considered enabled hazelcast diagnostics logging?  It is kind of verbose but hazelcast often asks for it when debugging issues.  If not enabled by default, we should expose an option to provide enable it easily.

charts/pega/templates/pega-environment-config.yaml
Saurabh-16: As per suggestions from Dave and you , I am going with this name:-
HZ_SERVER_URL

charts/pega/values.yaml
Saurabh-16: yes, Few mandatory parameters of subchart values.yaml can be overridden in main chart values.yaml . The same technique has been applied to installer and pega-search subchart as well.

charts/pega/templates/_helpers.tpl
Saurabh-16: I moved the empty definition to hazelcast _helpers.tpl.

They would be defined in the platform infinity SCE specific helm templates . The annotations are for fnx depoyment , so we need to override the empty definition for CMC just like `generatedPodAnnotations`

charts/pega/charts/hazelcast/templates/_helpers.tpl
girirajguptagithub: Right it is 
/hazelcast/health/Cluster-State(status should be 200) and not the /hazelcast/health/cluster-safe

charts/pega/charts/hazelcast/values.yaml
nanil4568: Created an issue to expose options related to hazelcast diagnostics.

charts/pega/charts/hazelcast/templates/_helpers.tpl
Saurabh-16: ok...I changed it to cluster-state.

PraneethaYanala: It is a recommendation from hazelcast to have minimum 3 servers for a cluster, so we want HZ cluster to be formed before client connects to the cluster.

charts/pega/templates/pega-environment-config.yaml
```diff
@@ -54,5 +54,14 @@ data:
...
-
+{{ if .Values.hazelcast.enabled }}
+  HZ_DISCOVERY_K8S: "true"
+  HZ_CLIENT_MODE: "true"
+{{ if .Values.hazelcast.client.clusterName }}
+  # Hazelcast Cluster Name
+  HZ_CLUSTER_NAME: {{ .Values.hazelcast.client.clusterName }}
+{{- end }}
+  # DNS of hazelcast service
+  HZ_SERVER_URL: {{ template "hazelcastName" . }}-service.{{ .Release.Namespace }}.svc.cluster.local
```
pega-embate, line 65: This is not a URL it is a hostname.   I think this name is confusing. Comment should be updated as well.

Saurabh-16: ok...I missed updating the comment. Changing the comment and environment variable name.

charts/pega/charts/hazelcast/templates/_helpers.tpl
APegaDavis: Is it possible to avoid hard-coding endpoints and just take the full URL? An issue with Stream Service just came up last week where we had originally [hardcoded the endpoint](https://github.com/pegasystems/docker-pega-web-ready/blob/1bc9c990317294f35e3da06f08fb1945bf8b255e/tomcat-conf/Catalina/localhost/prweb.xml#L20) (taking in the base URL as a param), but they needed to change that. This now means we have to make changes to the helm charts/docker image. If we had used the full URL as the param we wouldn't have had to make any changes.

Saurabh-16: So in this case, we need to construct the URL dynamically using service name and namespace. Namespace would only be available at run time. Also this is at the helm level. So are you suggesting to have this endpoint defined in some local variable and then refer that local variable in if condition , instead of doing a check on partially hardcoded url.
We tested this helm charts using infinity SCE and didn't see any issues.

Also in the docker image changes , we are passing complete URL from helm config maps and we don't need to append any additional path params to it.  See this PR:-
https://github.com/pegasystems/docker-pega-web-ready/pull/83

APegaDavis: I think ideally we don't have to know about the details of these services configuration. Getting the namespace at deployment time makes sense, but if we could avoid the specifics of their endpoint in our code somehow I think that would be ideal. I'm not suggesting there's any problem with the way things work now, just trying to think of ways to avoid tight coupling between services. I think providing a default endpoint and allowing it to be overridden like you suggest sounds good.

charts/pega/charts/hazelcast/templates/_helpers.tpl
```diff
@@ -0,0 +1,38 @@
...
+
+{{- define "isHazelcastEnabled" }}
+ {{- if .Values.enabled -}}
+  true
+ {{- else -}}
+  false
+ {{- end -}}
+{{- end }}
+
+{{- define "generatedHazelcastServicePodAnnotations" }}
```
MadhuriArugula, line 13: Looks like this definition will be overridden in helm deployments other than CMC. Can we add a comment to be clear on how this definition will be used in similar lines to [pega/_helpers.tpl](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/templates/_helpers.tpl#L226)

charts/pega/charts/hazelcast/templates/pega-hz-environment-config.yaml
```diff
@@ -0,0 +1,36 @@
...
+{{- end }}
+{{- if .Values.server.hz_max_join_seconds }}
+  HZ_MAX_JOIN_SECONDS: {{ .Values.server.hz_max_join_seconds | quote }}
+{{- end }}
+{{- if .Values.server.group_name }}
+  GROUP_NAME: {{ .Values.server.group_name | quote }}
+{{- end }}
+{{- if .Values.server.mancenter_url }}
+  MANCENTER_URL: {{ .Values.server.mancenter_url | quote }}
+{{- end }}
```
MadhuriArugula, line 35: I'm not sure if this is handled within the image but do we want to define any defaults for any of these environment variables? for example, `HZ_OPERATION_THREAD_COUNT`

terratest/src/test/pega/pega-environment-config_test.go
```diff
@@ -29,6 +29,7 @@ func TestPegaEnvironmentConfig(t *testing.T){
 				SetValues: map[string]string{
 					"global.provider":        vendor,
 					"global.actions.execute": operation,
+					"hazelcast.enabled": "true",
```
MadhuriArugula, line 32: Since hazelcast is like an addon to the main chart, can this test be moved to a new file to isolate with the actual deployments. 
For example, constellation is a new addon to Pega, hence we have added its test cases in a new file. Please refer these two files for reference [1](https://github.com/pegasystems/pega-helm-charts/blob/master/terratest/src/test/pega/pega-tier-deployment-with-c11n_test.go) [2.](https://github.com/pegasystems/pega-helm-charts/blob/master/terratest/src/test/pega/pega-c11n-deployment_test.go)

charts/pega/charts/hazelcast/templates/_helpers.tpl
Saurabh-16: good point. I will add a comment.

charts/pega/charts/hazelcast/templates/pega-hz-environment-config.yaml
Saurabh-16: Actually not needed as these are not mandatory parameters for hazelcast to work and is up to end user to configure the hazelcast as per their requirement.

terratest/src/test/pega/pega-environment-config_test.go
Saurabh-16: Ok. I will do it. This way it looks clean but we are duplicating the deployments.

MadhuriArugula: It doesn't duplicate the code, you can assert only what is required in the new addon's. 
Rest of them will be evaluated as per the actual deployment YAML. Reference links might help you in how to achieve.

Saurabh-16: Yeah , checked it again. We are calling same deployment files for asserting web , batch and stream.

MadhuriArugula (approved):

PR #213, opened 2020-12-03
US-391911-1:changes to support hz c/s authentication

SushmaReddyLoka:
Changes to support hz c/s authentication. We take username and password, and create secrets out of this and mount volumes created out of these secrets on container files. Tomcat and hz server process gets these values through prweb.xm and env variables respectively.


charts/pega/values.yaml
```diff
@@ -227,3 +227,7 @@ hazelcast:
   enabled: false
   # No. of initial members to join
   replicas: 3
```
Saurabh-16, line 229: Can we have these fields only in sub-chart . We will just try to keep minimal params in Values.yaml file.

charts/pega/values.yaml
```diff
@@ -227,3 +227,7 @@ hazelcast:
...
+  # UserName to be used in client server hazelcast mode for authentication
+  username: ""
+  # Password to be used in client server hazelcast mode for authentication
+  password: ""
```
Saurabh-16, line 233: Can we have these fields only in sub-chart . We will just try to keep minimal params in Values.yaml file. 
Also in the comment , we need to add like these settings are applicable from 8.6+ versions

charts/pega/charts/hazelcast/templates/pega-hz-deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
   {{- if ( .Values.imagePullPolicy ) }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
   {{- end }}
+        volumeMounts:
+        - name: {{ template "pegaVolumeCredentials" }}
+          mountPath: "/opt/hazelcast/secrets"
```
Saurabh-16, line 30: Do we need to create this folder `/opt/hazelcast/secrets` inside the docker container in pega-web-ready image or it just works by default? Please check this reference.
https://github.com/pegasystems/docker-pega-web-ready/blob/master/Dockerfile#L58

charts/pega/templates/_helpers.tpl
```diff
@@ -236,3 +236,11 @@ true
...
     defaultMode: 420
 {{- end}}
+
+{{- define "isHzEnabled" }}
+  {{- if (eq .Values.hazelcast.enabled true) -}}
+    true
+  {{- else -}}
+    false
+  {{- end -}}
+{{- end }}
```
Saurabh-16, line 246: We already have this template in hazelcast subchart helpers.tpl . Is there any need to have it at this place as well.

charts/pega/charts/hazelcast/templates/pega-hz-deployment.yaml
nanil4568: opt/hazelcast/secrets folder is being created from docker-entrypoint.sh of clustering-service
https://git.pega.io/projects/PS/repos/clustering-service/pull-requests/9/diff#distribution/docker/src/main/resources/docker-entrypoint.sh?t=8

charts/pega/values.yaml
nanil4568: As discussed, since these fields are mandatory for hazelcast(without which the deployment fails), we have included them in the main chart's values.yaml

charts/pega/templates/_helpers.tpl
nanil4568: removed this template and modified the references

taz-pega-work (approved):
The client-facing comments in pega/values.yaml chart look great!

charts/pega/charts/hazelcast/templates/pega-hz-deployment.yaml
dcasavant: At the time the entrypoint runs, the secret will either be injected already (or not).

charts/pega/templates/_pega-credentials-secret.tpl
```diff
@@ -15,6 +15,12 @@ data:
   CASSANDRA_USERNAME: {{ .Values.dds.username | b64enc }}
   # Base64 encdoded password for connecting to cassandra
   CASSANDRA_PASSWORD: {{ .Values.dds.password | b64enc }}
+  {{ if $.Values.hazelcast.enabled }}
+  # Base64 encoded username used for authentication in hazelcast client server mode
+  HZ_CS_AUTH_USERNAME: {{ .Values.hazelcast.username | b64enc }}
+  # Base64 encoded password used for authentication in hazelcast client server mode
+  HZ_CS_AUTH_PASSWORD: {{ .Values.hazelcast.password | b64enc }}
+  {{ end }}
```
kishorv10, line 23: Does Pega node require these credentials? If not its worth considering creating separate secret for hazlecast credentials

charts/pega/charts/hazelcast/templates/pega-hz-deployment.yaml
kishorv10: I think volume mount happens before entrypoint. I doubt if creating folder in entrypoint has any affect. If user has required permissions then folders get automatically created during mount.

kishorv10 (approved):

Saurabh-16 (approved):

charts/pega/charts/hazelcast/templates/pega-hz-deployment.yaml
nanil4568: Will verify this by reverting the changes made in clustering-service repo

PR #214, opened 2020-12-03
option to include serviceAccountName in deployments

zitikay:
In FNX all deployments need `serviceAccountName: pega-infinity-sa` to communicate with backing-services. Currently no way to set that.

### DIFF
--- a/charts/pega/templates/_pega-deployment.tpl
+++ b/charts/pega/templates/_pega-deployment.tpl
@@ -44,6 +44,11 @@ spec:
 {{- include "generatedPodAnnotations" .root | indent 8 }}
 
     spec:
+{{- if .custom }}
+{{- if .custom.serviceAccountName }}
+      serviceAccountName: {{ .custom.serviceAccountName }}
+{{- end }}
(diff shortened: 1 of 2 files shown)


charts/pega/templates/_pega-deployment.tpl
```diff
@@ -44,6 +44,11 @@ spec:
 {{- include "generatedPodAnnotations" .root | indent 8 }}
 
     spec:
+{{- if .custom }}
```
APegaDavis, line 47: Could we just use 'serviceAccountName'?

misterdorito (approved):

charts/pega/templates/_pega-deployment.tpl
APegaDavis: Nm, looks like this is the convention.

APegaDavis (approved):

dcasavant (approved):

taz-pega-work (commented):
no wording changes needed. ;-)

PR #215, opened 2020-12-07
ISSUE-149 Update default securityContext and mount point for stream pods

MadhuriArugula:
These PR changes are made to resolve the  #149 

Kafka write location needs to be updated to ensure that the mount point in helm charts and kafka service topics/partitions write location are not same to avoid known issue of lost+found directory ches/docker-kafka#28

These changes are backward compatible as we are creating a sub folder under /opt/pega/kafkadata

Since, opt/pega/kafkadata is a mounted folder, its permission will be root:root and that will not have permissions to create or write data inside it. Therefore, securityContext has been introduced in the pod spec so that our default mounted volumes will also have correct permissions. 

Defaulting the securityContext inline with our Pega published docker images and documenting a way for the users to provide runAsUser with custom image is used in their deployments. 

Refer BUG-575909 for more details.

PR #216, opened 2020-12-08
Allow Openshift to use arbitrary user at runtime.

pegatim:
Disabling the use of runAsUser in search container for Openshift and allowing the use of arbitrary user at runtime.


MadhuriArugula:
This PR has changes that are already requested as part of #215

timlaipega:
Do want to add since this specific commit modified only pega-search-deployment.yaml so this mod would impact only the search pod.

Saurabh-16:
This PR has same changes as https://github.com/pegasystems/pega-helm-charts/pull/215 . Closing this PR.
#215

PR #217, opened 2020-12-18
US-392813 Make HPA targets independently optional in the HPA template

pmaslak92:
US-392813 Small fix

US-392813 Small fix

US-392813 Add automatic test


charts/pega/README.md
```diff
@@ -403,6 +403,8 @@ Parameter           | Description    | Default value
 `hpa.maxReplicas`   | Maximum number of replicas that HPA can scale-up  | `5`
 `hpa.targetAverageCPUUtilization` | Threshold value for scaling based on initial CPU request utilization (The default value is `70` which corresponds to 70% of 2) | `70`
 `hpa.targetAverageMemoryUtilization` | Threshold value for scaling based on initial memory utilization (The default value is `85` which corresponds to 85% of 6Gi ) | `85`
+`hpa.disableCPUTarget` | Set to true if you want to remove scaling based on CPU utilization and only rely on memory utilization | false
```
dcasavant, line 406: Rather than have the flag be to disable the target, can we have it be true=on and false=off and avoid the negative parameter?  We can still default to `true` if not specified.

charts/pega/values-large.yaml
```diff
@@ -214,6 +215,7 @@ global:
 
       hpa:
         enabled: true
```
dcasavant, line 217: Why does the stream tier have scaling set to true at all?  Should be false to match values.yaml right?

dcasavant (changes requested):
See inline comments

charts/pega/README.md
pmaslak92: @dcasavant I am not an expert in helm templates, but from what I see it doesn't distinguish between a property being absent or it being  equal to false. We wanted to make this backwards-compatible, so that if someone has an existing values file without the new properties, it works as before. We tried doing it like you suggest, but in the template when we used "(default true .hpa.enableCpuTarget)" it would always evaluate to true, even if we set this value explicitly to false, so it looks like this operator is just an OR.

Is there a different way to provide a default value that will work properly with booleans?

charts/pega/values-large.yaml
pmaslak92: @dcasavant I don't think this is a question for me, this story was about making it possible to remove one of the targets in HPA and do this for dataflow tiers. The stream tier is outside the scope of this story and it's a different team that owns it.

charts/pega/README.md
andreiadamian: @pmaslak92 I agree with @dcasavant - let's not use negatives. I believe the user story asks for .hpa.cpu and .hpa.memory. 
You use default function in a reversed order, it should be "(default .hpa.enableCpuTarget true)"

pmaslak92: @andreiadamian I have tried in this order too. You can see it here - https://github.com/pmaslak92/pega-helm-charts/commit/b5796756a6beefd3a1869c8bdf8d7c2615138ea1 - the test is failing, because helm templates don't distinguish between a property being absent or being equal to false.

As discussed with @andreiadamian, I changed the property to be a string "on" or "off". @dcasavant Can you take a look?

taz-pega-work: I believe the this wording must be updated...here's my guess at an appropriate alternative to line 406: "Set to on to enable scaling based on CPU utilization; set to off to enable scaling based on memory utilization | on"

pmaslak92: @taz-pega-work it was "on" and "off" at some point, but it was requested that this is a boolean,  so now it's 'true' and 'false' everywhere. The current wording is correct.

charts/pega/values.yaml
```diff
@@ -137,6 +137,7 @@ global:
 
       hpa:
         enabled: true
+        memoryTarget: false
```
dcasavant, line 140: Should be `enableMemoryTarget` right?

pmaslak92: You are right. I just pushed a commit which fixes that.

dcasavant (approved):
Looks

taz-pega-work (approved):

misterdorito (approved):

MadhuriArugula (approved):

PR #229, opened 2021-01-11
Added the recommended default JVM arguments as CATALINA_OPTS

yashwanth-pega:
For better tunability of JVM Pega recommends defaults for some of the JVM arguments. Any JVM arguments with the flexibility to override can be provided as **catalinaOpts** under a specified tier. Note that _some_ of the arguments don't have the flexibility to be overridden and hence are defaulted in Docker image. This is complemented by related changes for **CATALINA_OPTS** in https://github.com/pegasystems/docker-pega-web-ready image. 
Link to the mentioned PR: https://github.com/pegasystems/docker-pega-web-ready/pull/95

### DIFF
--- a/charts/pega/templates/_helpers.tpl
+++ b/charts/pega/templates/_helpers.tpl
@@ -132,7 +132,11 @@ until cqlsh -u {{ $cassandraUser | quote }} -p {{ $cassandraPassword | quote }} 
   value: "{{ .node.javaOpts }}"
 # Additional CATALINA arguments
 - name: CATALINA_OPTS
+{{- if .node.catalinaOpts }}
   value: "{{ .node.catalinaOpts }}"
+{{- else }}
+  value: ""
+{{- end }}
 # Initial JVM heap size, equivalent to -Xms
 - name: INITIAL_HEAP
 {{- if .node.initialHeap }}
--- a/charts/pega/templates/_helpers.tpl
+++ b/charts/pega/templates/_helpers.tpl
@@ -130,6 +130,9 @@ until cqlsh -u {{ $cassandraUser | quote }} -p {{ $cassandraPassword | quote }} 
 # Additional JVM arguments
 - name: JAVA_OPTS
   value: "{{ .node.javaOpts }}"
+# Additional CATALINA arguments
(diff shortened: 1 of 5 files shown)


charts/pega/templates/_helpers.tpl
```diff
@@ -130,6 +130,9 @@ until cqlsh -u {{ $cassandraUser | quote }} -p {{ $cassandraPassword | quote }}
 # Additional JVM arguments
 - name: JAVA_OPTS
   value: "{{ .node.javaOpts }}"
+# Additional CATALINA arguments
+- name: CATALINA_OPTS
+  value: "{{ .node.catalinaOpts }}"
```
MadhuriArugula, line 135: catalinaOpts is not present in actual values.yaml and might result in nil reference during deployment. Can we validate that or look for node.catalinaOpts before assigning to an ENV only if present.

MadhuriArugula (changes requested):

charts/pega/templates/_helpers.tpl
yashwanth-pega: Added the validation check for catalinaOpts

MadhuriArugula (approved):

NekhilKotha (approved):

pega-embate (approved):

yashwanth-pega:
@taz-pega-work  Thanks for your feedback. 
I've incorporated them. 
Regarding these:
Lines 42 - 44 non-bocking[**definitely a typo**]??? --> non-blocking? ; dev/random[**if you meant this, not a typo and is intended**] or dev/urandom - I think it's a typo
Line 39 ...why are you setting these if you would think of overwriting them somewhere else? the context isn't obvious. 
**Ans**: To be clear, what I want to convey here is: These JVM arguments are hardcoded with Recommended values in the Docker image. You are not allowed to set them explicitly, in the sense that if you pass them explicitly, they _will not be obeyed_.

dcasavant (approved):

taz-pega-work (approved):
looks great! Thanks for working with me to incorporate useful clarifications!

PR #230, opened 2021-01-11
added .custom.sidecarContainers to use sidecar Containers

mathijs55:
I've added .custom.sidecarContainers to be able to use sidecar containers. for us this is extremely helpfull to use with extracting logging, but i'm sure there would be other functions that can be helpful to others.


MadhuriArugula:
Can you add more details about the functionality and the purpose it solves? Also, if we are providing any custom configuration, provide documentation similar to [https://github.com/pegasystems/pega-helm-charts/blob/58c6af70e132d4fe58192bc5facf5ba804905629/charts/pega/README.md#environment-variables](url)

mathijs55:
I messed up my pull request with some wrong merges. I will open a new pull request with more details as Madhuri suggested.

Issue #234, opened 2021-01-14
pega/addons charts in version 1.4.0 are unavailable (returning 404 error)


rbogendoerfer:
**Describe the bug**
The pega/addons charts version 1.4.0 is not available in the repo. Despite being listed (i.e. when using 'helm search repo pega') trying to access the charts always fails with an error.

**To Reproduce**
The helm repo has been updated (helm repo update). After that accessing the latest pega/addons version 1.4.0 fails.

helm inspect:
$ helm inspect values pega/addons
Error: failed to download "pega/addons" (hint: running `helm repo update` may help)

helm pull:
$ helm pull pega/addons
Error: failed to fetch https://dl.bintray.com/pegasystems/pega-helm-charts/addons-1.4.0.tgz : 404 Not Found

Accessing other versions still works.

**Expected behavior**
The helm commands should work as expected, i.e. display the values or pull the charts from the repo.

**Chart version**
pega/addons version 1.4.0
No modifications, just using the repo https://dl.bintray.com/pegasystems/pega-helm-charts

**Server (if applicable, please complete the following information):**
n/a

**Additional context**
n/a

anujgarg06:
@MadhuriArugula @dcasavant It is still giving error for addon file, seems like 1.4.1 version doesn't exist for it
helm search repo pega
NAME                    CHART VERSION   APP VERSION     DESCRIPTION
pega/pega               1.4.1                           Pega installation on kubernetes
pega/addons             1.4.1           1.0             A Helm chart for Kubernetes
pega/backingservices    1.4.1                           A Helm chart for Kubernetes

>> helm inspect values pega/addons > addons.yaml
Error: failed to download "pega/addons" (hint: running `helm repo update` may help)
>> helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "pega" chart repository
Update Complete. ⎈Happy Helming!⎈
>> helm inspect values pega/addons > addons.yaml
Error: failed to download "pega/addons" (hint: running `helm repo update` may help)

MadhuriArugula:
> @MadhuriArugula @dcasavant It is still giving error for addon file, seems like 1.4.1 version doesn't exist for it
> helm search repo pega
> NAME CHART VERSION APP VERSION DESCRIPTION
> pega/pega 1.4.1 Pega installation on kubernetes
> pega/addons 1.4.1 1.0 A Helm chart for Kubernetes
[...]

PR #235, opened 2021-01-14
Fix indexing of the backing services file name

dcasavant:
As reported in #234 we are no longer publishing the addons file.  It appears that this may be caused by `before_deploy.sh` applying the same index to both addons and then backingservices to replace the descriptor-template.json with the actual file names.

This PR attempts to fix the publishing of the addons chart by updating the before deploy script with a proper index for the backing services tgz.


MadhuriArugula (approved):

NekhilKotha (approved):

Heshma (approved):

MadhuriArugula:
@dcasavant Can we merge this and create a new release with same version for all the charts? Currently, addons latest version is 1.3.5 while others are at 1.4.1.

anujgarg06:
Hi, 
Do we have a workaround to overcome this error while your team is working on its fix?

pega-kshev (approved):

PR #239, opened 2021-01-26
added .custom.sidecarContainers to use sidecar Containers

mathijs55:
I've added a .custom.sidecarContainers annotation and also updated the readme file with an example. We use this sidecar container configuration to have a shared volume mount on which the pega container writes its logging. the rsyslog sidecar container then copies this logging and sends it to an ELK-stack which is hosted elsewhere.

### DIFF
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -466,6 +466,21 @@ tier:
       serviceAccountName: MY_SERVICE_ACCOUNT_NAME
 ```
 
+### Sidecar Containers
+
+If the pod needs to run with one or more [sidecar containers](https://kubernetes.io/docs/concepts/cluster-administration/logging/#using-a-sidecar-container-with-the-logging-agent), you can specify custom  'sidecarContainers' for your deployment tier.
+
+Example:
+
+```yaml
+tier:
+  - name: my-tier
+    custom:
+      sidecarContainers:
+      - name: SIDECAR_NAME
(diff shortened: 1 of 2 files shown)


MadhuriArugula:
@mathijs55 Can you add a test case for custom sidecar containers [here](https://github.com/pegasystems/pega-helm-charts/blob/master/terratest/src/test/pega/pega-tier-deployment_test.go)

mathijs55:
@MadhuriArugula  I'm sorry, I don't know how to write a test-case for this. I don't know what language that is...

MadhuriArugula:
> @MadhuriArugula I'm sorry, I don't know how to write a test-case for this. I don't know what language that is...

@mathijs55 Test cases are written in GO lang, you can refer to the example test [here](https://github.com/pegasystems/pega-helm-charts/blob/master/terratest/src/test/pega/pega-tier-deployment_test.go)

marpada:
@mathijs55 thank you for doing this! Hopefully it will get some love from maintainers as it is almost a "must" for deploying Pega in prod.

If I can make a suggestion I think it would be better to add the sidecar containers after the pega-web-tomcat container to maintain some existing like 'kubectl exec', which by default executes in the first container.

pega-talba:
This PR is being adopted in https://github.com/pegasystems/pega-helm-charts/pull/371.

Thank you for your contributions! 🙂

Issue #240, opened 2021-01-26
Turning off Reverse DNS lookup 


raghavf1:
**Describe the bug**
I would like to disable reverse DNS lookups in PRPC platform to reduce amount of configuration required for supporting reverse DNS and also to minimize the impact it has on performance of requests served.

WebStandardImpl.makeEtierRequest(IWebStandard, HttpServletRequest, HttpServletResponse, Map, Object, DirectStreamWriter) has an intelligent way to fetch remote host only the first time when the session is established. It stores the host name in the http session when the session is established for the first time then it uses the value from session rather than resolving from request.

The issue is the value retrieved by makeEtierRequest is not being used at all the places in engine and resolving from request at other places. We can make change to retrieve the host value from session everywhere.

**To Reproduce**
This is non-trivial to reproduce

**Expected behavior**
n/a

**Chart version**
n/a

**Server (if applicable, please complete the following information):**
n/a

**Additional context**
For internal reference, also see US-139237,BUG-256208

APegaDavis:
@raghavf1 going to close this because it's inactive and the PR you opened was recently closed for similar reasons.

PR #241, opened 2021-01-26
ISSUE-240:Turning off Reverse DNS lookup

raghavf1:
Reverse DNS is used to lookup the hostname of the IP address where the request is originating from. This is typically used for logging purposes, so domain names can be shown instead of IP addresses. However, this doesn't make sense in cloud and/or cluster deployments, since the originating IP address will be an internal address, e.g. the address of the load balancer etc. Since doing a reverse DNS will not provide any value, it will result in unneeded overhead.
 Reverse DNS lookups have also been shown to be a source of spikes, in case the DNS server / resolver get into temporarily non-responsive.
Regarding backward compatibility: Since its main purpose is logging, there shoudn't be any b/w compatibility issues.

### DIFF
--- a/charts/pega/config/deploy/prconfig.xml
+++ b/charts/pega/config/deploy/prconfig.xml
@@ -7,6 +7,7 @@
         <env name="database/databases/PegaDATA/dataSource" value="java:comp/env/jdbc/PegaRULES"/>
         <env name="security/urlaccesslog" value="NORMAL" />
         <env name="security/urlaccessmode" value="WARN" />
+        <env name="http/reversednslookup" value="false" />​
         <!-- Most nodes have a 'default' classification and for these nodes, no additional changes need to be made to this file.  However,
         if this is node has a non-general purpose, for example: 'Agent', then the node classification setting should be added to this file. -->
         <!--env name="initialization/nodeclassification" value="Agent" /  -->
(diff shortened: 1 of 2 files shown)


MadhuriArugula:
@raghavf1 Can you update the PR with an explanation of the change and that it will not impact backwards compatibility?
Also you might want to fix the unit tests failed because of this change.

raghavf1:
> @raghavf1 Can you update the PR with an explanation of the change and that it will not impact backwards compatibility?
> Also you might want to fix the unit tests failed because of this change.

@MadhuriArugula I have updated description of the PR and fixed the corresponding tests. Please let me know if you need any further info

charts/pega/config/deploy/prconfig.xml
```diff
@@ -7,6 +7,7 @@
         <env name="database/databases/PegaDATA/dataSource" value="java:comp/env/jdbc/PegaRULES"/>
         <env name="security/urlaccesslog" value="NORMAL" />
         <env name="security/urlaccessmode" value="WARN" />
+        <env name="http/reversednslookup" value="false" />​
```
MadhuriArugula, line 10: @raghavf1 Is it mandatory to set reverse DNS lookup to` false `by default? or can this be configurable from values.yaml? 
As per the user story and bug item mentioned in #240, it is under Engine backlog. 

Can you confirm if this was verified with Engine teams? Some approvals from Engine can help us proceeding with merge.

kishorv10:
Since there has been no activity on this pull request for the past year, I will proceed to close it. Please feel free to reopen it if necessary.

PR #242, opened 2021-01-28
BUG-591266: add deployment name and rename tier-based objects.

misterdorito:
Adding the ability to set an 'deployment' name.  This environment name will be prepended to the tier names replacing "pega".  This will impact the naming of all of the corresponding tier-centric objects (deployments, statefulsets, services, pods, ingresses, etc.).

The particular motivation for adding this is that currently the PDC service is unable to differentiate stream nodes within a given PDC tenant as all of the pod names are pega-stream-0 and pega-stream-1. 

To leverage this ability, set global.deployment.name to the desired prefix.  The prefix must consist of an DNS friendly identifier as it will be incorporated into a number of k8s object (which have this restriction).  An example of a suitable value would be: "app1-dev".
 
The default if not specified will remain "pega".


charts/pega/README.md
```diff
@@ -130,6 +130,30 @@ docker:
     imagePullPolicy: "Always"
 ```
 
+## Environment Name (Optional)
+
+Set the environment name to help differentiate this deployment from others.  The provided name will be used as a prefix to the various Pega tiers (and the k8s objects created on their behalf).  The environment name must consist of a DNS friendly name as it will be incorporated into the names of k8s objects. For example:
+```yaml
+global:
+  environment:
+    name: app1-dev-
```
pega-embate, line 139: Is this tied into the global.domain parameter that is used for the hostname for this environment?  It seems like it would be less confusing if the environment name value were derived from that value.

misterdorito: No.

I am not familiar with global.domain (only global.tier[].ingress.domain).  Can you point me at a reference?

pega-embate: Sorry - I meant global.dnsname.  I believe this is used for configuration for the ingress.
That said, I also noticed that there is a global.ingress.domain param.  Can all of these be consolidated?  It seems like they all could potentially be set to the same value.

misterdorito: I couldn't find a reference to 'dnsname' grepping through the project.

I do see a couple of 'domain' args:

- global.tier[].ingress.domain 
- global.tier[].service.domain
- global.tier[].domain

Though I'm inclined to say that we should not derive this from the domain used for ingress.  My thought is that these names would be internal identifiers that don't necessarily relate to their public domain names.  My intention in bringing up DNS naming conventions was to indicate the constraints on the names given:

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

(Don't want folks thinking they can include spaces and such.)

misterdorito:
> Should there be a `-` at the end? Or is that automatically incorporated?

Right now, yes there should be a '-' (global.environment.name: app1-dev-) the way it is coded currently.  Happy to do either.  Do you have a preference for adding the dash programmatically?

misterdorito:
> Should we just apply this to search as well for consistency? Any reason not to do it?

I haven't tried, but it shouldn't be a problem.

charts/pega/README.md
pega-embate: global.dnsname is referenced in the sce.  @zitikay do you have any thoughts about this?

misterdorito:
> wordsmithing for the description...that is an ACE example! nice job! ;-)
> -->Specify an environment name that the deployment uses as a prefix to help differentiate this deployment in your environment. The deployment adds this prefix to the various Pega tiers and the associated k8s objects in your deployment. Your environment name must be a DNS-friendly name so the deployment can successfully add it to k8s object names. For example:

So, based on your wordsmithing above, I think it would be appropriate to change global.environment.name to global.deployment.name (reading you words above we're using environment in 2 different ways where the first one is really describing a deployment).  With that in mind:

Specify a deployment name that is used to differentiate it in your environment.  This name will be prepended  to the various Pega tiers and the associated k8s objects in your deployment.  Your deployment name should be constrained to lowercase alphanumeric and '-' characters.

Thoughts?

charts/pega/README.md
pega-embate: Actually, @misterdorito you added the dnsname param (in the SCE).  I'm just trying to avoid the confusion of having similar parameters with slightly different uses.  

How would you differentiate the following to a user?
global.tier[].ingress.domain
global.dnsname
global.environment.name

misterdorito: global.dnsname: This is only used in the SCE and is largely need to work around the limited utility of the SCE tooling.  It is internal to the SCE and among the things that I'd like to refactor out of existence -- as will probably happen soon given my understand of work to come in the near future (as there is need for multiple ingresses for different  types of deployments.

I'd rather we did not derive these 'deployment.name'-s (see other comments below regarding the change of nomenclature) from ingress/dns names as I'd expect that these will be relatively short names use to distinguish deployments (mostly so that PDC is capable of reporting on stream nodes in client-managed cloud scenarios).  I'd expect this not to be widely used if PDC is not a concern.  For SCE deployments, this will not be used.  And as there are no guarantees that multiple ingresses in the same deployment will have commonality in their domains.

kishorv10 (commented):
@misterdorito  

- Are you not planning to add this entry in values.yaml?
- Should we also append the env name to `pega-credentials-secret.yaml`, `pega-registry-secret.yaml`, `pega-environment-config.yaml` object names? Though this is not need for original requirement of this bug, but doing this would make things consistent. It also opens  feature to deploy more than one pega helm releases in a single namespace
- Should we deal with the `-` bit programmatically. As I understand, if environment name not ended with `-` would lead some cluttered output, very minor thing to care for.

misterdorito:
> Should we just apply this to search as well for consistency? Any reason not to do it?

@dcasavant Looking a little more closely...

We could change this, but there is a minor wrinkle....  We'll have to change some the internal logic regarding isExternalSearch because some of this logic is based on the correlation between the pega-search service and the pegasearch.externalURL being hardcoded to http://pega-search in the values yaml.

Should be doable

shashikant-koder:
Hi @misterdorito ,  
Did we test RollingUpdates scenario ,  as rolling updates command use tier name.  ?

misterdorito:
@kishorv10 
> @misterdorito
> 
> * Are you not planning to add this entry in values.yaml?
> * Should we also append the env name to `pega-credentials-secret.yaml`, `pega-registry-secret.yaml`, `pega-environment-config.yaml` object names? Though this is not need for original requirement of this bug, but doing this would make things consistent. It also opens  feature to deploy more than one pega helm releases in a single namespace
> * Should we deal with the `-` bit programmatically. As I understand, if environment name not ended with `-` would lead some cluttered output, very minor thing to care for.

I've added global.deployment.name to values.yaml, and values-large.yaml.  Wasn't planning on adding it to values-minimal.yaml, but I have no objection to adding it.

 `pega-credentials-secret.yaml`, `pega-registry-secret.yaml`, `pega-environment-config.yaml`: added.

The charts are adding the dashes at this point.

misterdorito:
> Hi @misterdorito ,
> Did we test RollingUpdates scenario , as rolling updates command use tier name. ?

I haven't yet, but will take a look tomorrow.  I may have questions for you.

misterdorito:
@shashikant-koder 
> > Hi @misterdorito ,
> > Did we test RollingUpdates scenario , as rolling updates command use tier name. ?
> 
> I haven't yet, but will take a look tomorrow. I may have questions for you.

I have made changes to this PR to address this concern.  I have tested them manually (helm template ... ) and they look correct to me.  Please take a look.

misterdorito:
@dcasavant Do we want to rename the hazelcast service and constellation  services as well?

pega-embate:
@misterdorito Is the new deployment name correlated with the namespace at all?  I ask this because it sounds like PDC is using the namespace as a first lookup point to identify pods associated with a given customer (when problems are detected).  I'm just trying to avoid having  too many different ways to configure the name of an infinity deployment (looking at this from a support standpoint)

misterdorito:
> @misterdorito Is the new deployment name correlated with the namespace at all? I ask this because it sounds like PDC is using the namespace as a first lookup point to identify pods associated with a given customer (when problems are detected). I'm just trying to avoid having too many different ways to configure the name of an infinity deployment (looking at this from a support standpoint)

Is deployment name correlated to namespace? No.

My understanding of the PDC is that within a given PDC tenant, hostnames need to be unique.  I don't think it has any understanding of k8s namespaces at this point.  In the past unique system names would have caused the stream nodes to be an issue, but according to the PDC folks there is a range of platform releases where info included in the alert format doesn't include system name.  The PDC/stream tier uniqueness problem is only a CMC issue.  

More flavor... Renaming the stream tier could be achieved by renaming the tier name for stream, but the PDC folks posed objections because some clients insisted on using the same values.yaml for multiple deployments of the platform.  I can only surmise that they're passing in db connection info via cmd line args to helm.

pegaautomationuser:
Starting PR-242 validation on -> integ-all

pegaautomationuser:
Starting PR-242 validation on -> integ-all

charts/pega/charts/installer/templates/_helpers.tpl
```diff
@@ -54,6 +56,8 @@
 {{- end }}
 
 {{- define "waitForRollingUpdates" -}}
+{{- $deploymentName := printf "%s-" (include "installerDeploymentName" $) -}}
+{{- $deploymentNameRegex := printf "%s- " (include "installerDeploymentName" $) -}}
```
pega-talba, line 60: Should this also include the `-` as it is used where `pega-` was previously used?

terratest/src/test/pega/utilities.go
```diff
@@ -56,12 +57,14 @@ func VerifyInitContinerData(t *testing.T, containers []k8score.Container, option
 			require.Equal(t, []string{"job", "pega-db-install"}, container.Args)
 		} else if name == "wait-for-pegasearch" {
 			require.Equal(t, "busybox:1.31.0", container.Image)
-			require.Equal(t, []string{"sh", "-c", "until $(wget -q -S --spider --timeout=2 -O /dev/null http://pega-search); do echo Waiting for search to become live...; sleep 10; done;"}, container.Command)
+			require.Equal(t, []string{"sh", "-c", "until $(wget -q -S --spider --timeout=2 -O /dev/null http://" + depName + "-search); do echo Waiting for search to become live...; sleep 10; done;"}, container.Command)
 		} else if name == "wait-for-cassandra" {
 			require.Equal(t, "cassandra:3.11.3", container.Image)
+			//todo: update with depName?
```
pega-talba, line 63: Is this a thing we still want to do in this PR?

charts/pega/charts/installer/templates/_helpers.tpl
pega-talba: Discussed privately - the dash gets added elsewhere as part of the deployment names, and the key feature of `deploymentNameRegex` is that it includes the extra space.

terratest/src/test/pega/utilities.go
pega-talba: Discussed privately - This actually does work properly with cassandra, so the todo comments will be removed.

pega-talba (approved):

pegaautomationuser:
Starting PR-242 validation on -> integ-all

pegaautomationuser:
Starting PR-242 validation on -> integ-all

misterdorito:
I have performed an out-of-place upgrade with the changes in this branch and it has worked as expected.  None of the changes in this PR should impact an in-place upgrade.

charts/pega/README.md
```diff
@@ -133,6 +133,32 @@ docker:
...
+app1-dev-stream-0                 1/1     Running   0          24m
+app1-dev-stream-1                 1/1     Running   0          18m
+app1-dev-web-788cfb8cc4-6c5nz     1/1     Running   0          8m57s
+app1-dev-web-788cfb8cc4-gcltx     1/1     Running   0          24m
+```
+The default value is "pega" if it is unset.
+
 ## Tiers of a Pega deployment
 
 Pega supports deployment using a multi-tier architecture to separate processing and functions. Isolating processing in its own tier also allows for unique deployment configuration such as its own prconfig, resource allocations, or scaling characteristics.  Use the `tier` section in the helm chart to specify which tiers you wish to deploy and their logical tasks.
```
taz-pega-work, line 164: A little wordsmithing to make it slightly more direct:
Pega supports deployments that a multi-tier architecture model that separates application processing from k8s functions. Isolating processing in its own tier supports unique deployment configurations, including the Pega application prconfig, resource allocations, and scaling characteristics.  Use the `tier` section in the helm chart to specify which tiers you wish to deploy the tier with nodes dedicated to the logical tasks of the tier.

charts/pega/charts/pegasearch/templates/_helpers.tpl
```diff
@@ -1,13 +1,37 @@
...
+{{- $defaultSearchURL -}}
+{{- else -}}
+{{- $searchURLOverride -}}
+{{- end -}}
+{{- else -}}
+{{- $defaultSearchURL -}}
+{{- end -}}
+{{- end -}}
+
 {{- /* From the perspective of the Pega Infinity nodes, the search instances are always external when deployed via these charts,
```
taz-pega-work, line 26: tweaks:
From the perspective of the Pega Infinity nodes, Pega deployments always assume that the search instances are external to the deployment when you use these charts; however this check determines if search is deployed externally as defined in the backing search service configuration or if it is defined in the Pega-provided Docker search image.*/

charts/pega/charts/pegasearch/values.yaml
```diff
@@ -6,7 +6,8 @@ replicas: 1
 # node is set to the minimum number of master-eligible nodes that must be visible in order to form a cluster.
 # Configure this value using the formula (n/2) + 1 where n is replica count or desired capacity.
 minimumMasterNodes: 1
-# If externalURL is set, no search nodes will be deployed automatically, and Pega will use this search node url.
```
taz-pega-work, line 9: lines 8 & 10: ...no search nodes are deployed automatically, and Pega uses this search node url

charts/pega/charts/pegasearch/values.yaml
```diff
@@ -6,7 +6,8 @@ replicas: 1
...
+# If externalURL is set to a value different from the name of the k8s service defined in pega-search-service.yaml,
+# no search nodes will be deployed automatically, and Pega will use this search node url.
 externalURL: "http://pega-search"
 # Enter the docker image used to deploy Elasticsearch. This value will be ignored if using an external url.
 # Push the Elasticsearch image to your internal docker registry. This must be the same registry as the docker section above.
```
taz-pega-work, line 13: Push the Elasticsearch image to the same internal Docker registry that you specified in the Docker section above.

charts/pega/templates/_helpers.tpl
```diff
@@ -239,7 +255,18 @@ true
...
     # This name will be referred in the volume mounts kind.
-    secretName: {{ template "pegaCredentialsSecret" }}
+    secretName: {{ template "pegaCredentialsSecret" $ }}
     # Used to specify permissions on files within the volume.
     defaultMode: 420
 {{- end}}
+
+{{- define "deploymentName" }}{{ $deploymentNamePrefix := "pega" }}{{ if (.Values.global.deployment) }}{{ if (.Values.global.deployment.name) }}{{ $deploymentNamePrefix = .Values.global.deployment.name }}{{ end }}{{ end }}{{ $deploymentNamePrefix }}{{- end }}
+
+#As the pegasearch chart includes an externalURL value that is hardcoded to the name of the search service
```
taz-pega-work, line 265: I'm not sure how this challenge observation helps...do you need to make a helpful, definitive statement or can you drop this comment? It's not clear what you are declaring other than a future problem if you need to rename the search service. Perhaps state the Pega recommendation and leave it that it can't be changed?

taz-pega-work (commented):
I found a few areas that could be tweaked to improve the guidance.

pegaautomationuser:
Starting PR-242 validation on -> integ-all

pegaautomationuser:
Starting PR-242 validation on -> integ-all

MadhuriArugula (approved):

khick77 (approved):

pegaautomationuser:
Starting PR-242 validation on -> integ-all

pegaautomationuser:
Starting PR-242 validation on -> integ-all

pegaautomationuser:
Starting PR-242 validation on -> integ-all

pegaautomationuser:
Starting PR-242 validation on -> integ-all

charoensri:
Search POD starts up with error
[2021-05-28T02:10:35,125][WARN ][o.e.d.z.UnicastZenPing   ] [3luSi9Q] failed to resolve host [pega-search-transport]
java.net.UnknownHostException: pega-search-transport: Name or service not known
        at java.net.Inet4AddressImpl.lookupAllHostAddr(Native Method) ~[?:?]
 After I removed the "-transport" from the Search Deployment yaml, the issue is gone.  Would this be the right fix?
   env:
        - name: HOST_LIST
          ###Seri
          ###value: {{ template "searchName" . }}-transport
          value: {{ template "searchName" . }}
        - name: ES_JAVA_OPTS
          value: "-Xmx2g -Xms2g"
        - name: UNICAST_HOSTS
          ###Seri
          ###value: {{ template "searchName" . }}-transport
          value: {{ template "searchName" . }}

PR #245, opened 2021-02-11
Feature/us 400386

saikarthik528:
As part of hazelcast upgrade from version 3 to version 4, platform expects a prconfig flag (cluster/hazelcast/v4/enabled), to be set as true, which denotes that version 4 is enabled by default for any installations from pega-helm-charts.


MadhuriArugula (approved):

kishorv10 (approved):

charts/pega/config/deploy/prconfig.xml
```diff
@@ -15,4 +15,6 @@
         for this node.  This is useful for settings which are specific to this node and should not be shared by multiple nodes on this
         system. -->
 
+        <env name="prconfig/cluster/hazelcast/v4/enabled" value="true" type="java.lang.String"/>
+
```
Heshma, line 19: @saikarthik528 Could you please add a short comment on the usage of this setting?

charts/pega/charts/installer/config/prconfig.xml.tmpl
```diff
@@ -45,4 +45,5 @@
   <env name="install/parallel/threadcount" value="-1" />
   <env name="install/parallel/index/threadcount" value="-1" />
   <env name="compiler/javasourcelevel" value="1.7" />
+  <env name="prconfig/cluster/hazelcast/v4/enabled" value="true" type="java.lang.String"/>
```
PraneethaYanala, line 48: If we are setting in prconfig.xml then we should not add prconfig to config name it should be <env name="cluster/hazelcast/v4/enabled" value="true" type="java.lang.String"/>, same applies for all the prconfig files

PraneethaYanala, line 48: If we are setting in prconfig.xml then we should not add prconfig to config name it should be <env name="cluster/hazelcast/v4/enabled" value="true" type="java.lang.String"/>, same applies for all the prconfig files

saikarthik528: Addressed this in the latest push

charts/pega/config/deploy/prconfig.xml
saikarthik528: added comment about usage

PR #248, opened 2021-02-18
Add opt-in ReadOnly datasource to context.xml template

zitikay:
RO datasource is added with presence of new global.jdbc.readerurl value, and configured with RO env var credentials. This is to support [QueryRunner](https://community.pega.com/knowledgebase/articles/pega-cloud/query-runner-landing-page-testing-select-statements)

### DIFF
--- a/charts/pega/templates/pega-environment-config.yaml
+++ b/charts/pega/templates/pega-environment-config.yaml
@@ -10,6 +10,10 @@ data:
   DB_TYPE: {{ .Values.global.jdbc.dbType }}
   # JDBC URL of the DB where Pega is installed
   JDBC_URL: {{ .Values.global.jdbc.url }}
+{{- if .Values.global.jdbc.readerurl }}
+  # URI that the JDBC driver can be downloaded from
+  JDBC_RO_URL: {{ .Values.global.jdbc.readerurl }}
+{{- end }}
   # Class name of the DB's JDBC driver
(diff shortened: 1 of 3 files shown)


charts/pega/templates/pega-environment-config.yaml
```diff
@@ -10,6 +10,10 @@ data:
   DB_TYPE: {{ .Values.global.jdbc.dbType }}
   # JDBC URL of the DB where Pega is installed
   JDBC_URL: {{ .Values.global.jdbc.url }}
+{{- if .Values.global.jdbc.readerurl }}
+  # URI that the JDBC driver can be downloaded from
```
APegaDavis, line 14: Looks like a copy/paste remnant -- this is the read only URL not the driver UIR, right?

zitikay: Good catch, fixed this

khick77 (approved):

pega-talba (approved):

dcasavant (approved):

zitikay:
Merged:
For ReadOnly datasource can pass in `global.jdbc.readerurl` to values.yaml, and bring in DB_RO_USERNAME/PASSWORD as env vars. Additionally CONTEXT_XML_SNIPPET can be used for context.xml injection (rather than overriding whole file)

PR #260, opened 2021-03-15
changes for new upgrade types

shashikant-koder:
This PR to add new upgrade types like ZDT, Custom, OOPR and OOPD.


taz-pega-work (changes requested):
just marking this as deprecated like this can confuse clients..how about somethinglike this?
`out-of-place`  | DEPRECATED: If upgrading from 8.4 or later, use a 'zero-downtime' update type to complete an upgrade while continuing to work in your application. The 'out-of-place' upgrade minimizes downtime, but still requires some downtime.  This upgrade type places the rules into a read-only state, migrates the rules to your designated "new rules schema", migrates the data to your designated "temporary data schema", and then upgrades all of the rules to the new version. With the new rules upgraded, it shuts down the application briefly and upgrades the data in the temporary data schema, and then redeploys your application using the new rules.
TAZ>>
`zero-downtime`  | A zero-downtime upgrade synchronizes the required update steps to minimize downtime and you and your customers can continue to access and use their applications in your environment with minimal disruption. This upgrade type places the rules into a read-only state, migrates the rules to your designated "new rules schema", migrates the data to your designated "temporary data schema", and then upgrades all of the rules to the new version. After upgrading your rules schema to the new rules, the process performs a rolling reboot of your nodes and then upgrades the data in the temporary data schema. Finally, the process redeploys the application using the new rules.
I'm not sure where custom or ZDT-patch options are mentioned, but we should have them in here too, I think.

charts/pega/charts/installer/templates/_pega-installer-job.tpl
```diff
@@ -6,7 +6,7 @@ metadata:
   name: {{ .name }}
   namespace: {{ .root.Release.Namespace }}
   annotations:
-{{- if and .root.Values.waitForJobCompletion (or (eq .root.Values.global.actions.execute "install") (eq .root.Values.global.actions.execute "upgrade")) }}
+{{- if and (eq .root.Values.waitForJobCompletion "true") (or (eq .root.Values.global.actions.execute "install") (eq .root.Values.global.actions.execute "upgrade")) }}
```
arvasrikanth, line 9: This change is required for helm command to wait until the install/upgrade job to be completed. Also this is required only incase of cloudk

taz-pega-work (approved):

charts/pega/README.md
```diff
@@ -702,9 +702,12 @@ To **apply a Pega Platform patch** with zero downtime to your existing Pega plat
...
+`out-of-place` | `DEPRECATED`: If upgrading from 8.4.2 or later, use a 'zero-downtime' update type to complete an upgrade while continuing to work in your application. The 'out-of-place' upgrade minimizes downtime, but still requires some downtime. This upgrade type places the rules into a read-only state, migrates the rules to your designated "new rules schema", migrates the data to your designated "temporary data schema", and then upgrades all of the rules to the new version. With the new rules upgraded, it shuts down the application briefly and upgrades the data in the temporary data schema, and then redeploys your application using the new rules.
```
taz-pega-work, line 705: The positive cases work here, but you've introduced a problem in that clients running below 8.4.2 have no upgrade option...I believe the fix is to mark as deprecated and use the OOP choice as the choice for 8.4.1 and earlier; otherwise clients running 8.3 are likely going to need help deciding that they have to use the deprecated choice. Leave ZDT required versions as you have it, but I think you have to introduce OOP as:
"If upgrading from 8.4.1 or earlier, use this option to complete an upgrade." This type of upgrade minimizes downtime, but still requires some downtime.

charts/pega/README.md
```diff
@@ -702,9 +702,12 @@ To **apply a Pega Platform patch** with zero downtime to your existing Pega plat
...
+`zero-downtime` |  If upgrading from 8.4.2 and later, use 'zero-downtime' upgrade type. A zero-downtime upgrade synchronizes the required update steps to minimize downtime and you and your customers can continue to access and use their applications in your environment with minimal disruption. This upgrade type places the rules into a read-only state, migrates the rules to your designated "new rules schema", migrates the data to your designated "temporary data schema", and then upgrades all of the rules to the new version. After upgrading your rules schema to the new rules, the process performs a rolling reboot of your nodes and then upgrades the data in the temporary data schema. Finally, the process redeploys the application using the new rules.
```
taz-pega-work, line 706: "If upgrading from 8.4.2 and later, use 'zero-downtime' upgrade type." -->
"If upgrading from 8.4.2 and later, use this option."

charts/pega/README.md
```diff
@@ -702,9 +702,12 @@ To **apply a Pega Platform patch** with zero downtime to your existing Pega plat
...
 `custom` | custom upgrade can be used to configure upgrade in steps manner. User can specify upgrade steps : `enable_cluster_upgrade` `rules_migration` `rules_upgrade` `data_upgrade` `disable_cluster_upgrade` and it will perform configured steps.
+`out-of-place-rules` | An 'out of place rules' upgrade will perform rules upgrade on new rules schema.
```
taz-pega-work, line 708: An 'out of place rules' upgrade will perform rules upgrade on new rules schema.
An 'out of place data' upgrade will perform rules upgrade on new rules schema.
-->
An 'out of place' upgrade process performs the rules upgrade using the new rules schema and the data upgrade using the new data schema.
Question...aren't they creating a new data schema named "temporary data", not "new data"? the on-premises patches process does this: https://community.pega.com/knowledgebase/articles/keeping-current-pega/85/applying-patch-without-downtime this is what I said to do in my upgrade runbooks as well.

taz-pega-work (commented):
minor tweaks to clarify a few items.

charts/pega/README.md
shashikant-koder: Customer below 8.4.2 they can go with "out-of-place-rules" and "out-of-place-data" upgrade. Even they can use custom upgrade type if we pass upgrade_step -  `rules_migration` `rules_upgrade` `data_upgrade`  and they can try "in-place" upgrade too.

charts/pega/README.md
taz-pega-work: ok I think I see now...try this change to see if we capture the entire change required for any client:
`out-of-place` | `DEPRECATED`: If upgrading from 8.4.2 or later, use a 'zero-downtime' update type to complete an upgrade while continuing to work in your application; if upgrading from 8.4.1 or earlier, use any of the other supported upgrade types.
zero-downtime` |  If upgrading from 8.4.2 and later, use <<this option to complete an upgrade while continuing to work in your application.>> A zero-downtime upgrade...the rest looks good for this choice!
`custom` | Use this option for any upgrade in which you complete portions of the upgrade process in steps. Supported upgrade steps are: `enable_cluster_upgrade` `rules_migration` `rules_upgrade` `data_upgrade` `disable_cluster_upgrade`. To specify which steps to include in your custom upgrade, include them in your values.yaml file within the custom text. For example, to complete a rules upgrade without a data upgrade, you specify: `installer.upgradeType: "custom `enable_cluster_upgrade` `rules_migration` `rules_upgrade` `disable_cluster_upgrade`"`

`out-of-place-rules` | Use this option to complete an out-of-place upgrade of rules that the upgrade process migrates to the new rules schema. This schema will become the rules schema after your upgrade is complete.

`out-of-place-data` | Use this option to complete an out-of-place upgrade of data the upgrade process migrates to a new, temporary data schema. The upgrade process removes this temporary schema after your application is running with updated data.

taz-pega-work (commented):
So I don't think we can allow the new upgrade processes to stand alone without an example of each. Can we call out examples for a custom upgrade use case and at least one use case for the out-of-place rules? For the latter, I currently don't understand if we support doing an out-of-place rules upgrade alone without a data upgrade? Clients need explicit guidance.

kishorv10 (approved):

MadhuriArugula (approved):

PR #261, opened 2021-03-15
Allow server.xml and web.xml overrides in tier configmaps

APegaDavis:
* Allow overrides of Tomcat's server.xml and web.xml
* server.xml default copy included from config/deploy if no override is provided
* No web.xml default is provided, but will be picked up if a copy is placed in config/deploy
* web.xml is not included if no override is provided and the file does not exist in config/deploy

### DIFF
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -491,7 +491,7 @@ Parameter     | Description    | Default value
 `prlog4j2`    | A complete prlog4j2.xml file to inject.  | See [prlog4j2.xml](config/deploy/prlog4j2.xml).
 `contextXML`  | A complete context.xml template file to inject.  | See [context.xml.tmpl](config/deploy/context.xml.tmpl).
 `serverXML`   | A complete server.xml file to inject  | See [server.xml](config/deploy/server.xml).
-`webXML`   | A complete web.xml file to inject  | Not overridden by default.
+`webXML`      | A complete web.xml file to inject  | No default provided, but if `config/deploy/web.xml` exists, it will be used as the default.
 
 
 Example:
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -491,7 +491,7 @@ Parameter     | Description    | Default value
 `prlog4j2`    | A complete prlog4j2.xml file to inject.  | See [prlog4j2.xml](config/deploy/prlog4j2.xml).
 `contextXML`  | A complete context.xml template file to inject.  | See [context.xml.tmpl](config/deploy/context.xml.tmpl).
 `serverXML`   | A complete server.xml file to inject  | See [server.xml](config/deploy/server.xml).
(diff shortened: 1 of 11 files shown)


pega-embate:
Why are we allowing override of these config files?  Isn't the point of the prweb config to specify params that we want to have complete control of in k8s deployments?

pega-embate:
Assuming server.xml is still in the base image as well, how can we ensure they are kept in synch?

APegaDavis:
> Why are we allowing override of these config files? Isn't the point of the prweb config to specify params that we want to have complete control of in k8s deployments?

My mistake -- we want to allow override of server.xml and web.xml (not prweb.xml). I'll update my PR.

The goal is to allow customization of these files without having to provide additional enhancements every time a new category of customization is discovered, which has been a problem in the past.

> Assuming server.xml is still in the base image as well, how can we ensure they are kept in synch?

Since the base image is specific to a particular platform version it might make sense to do the same for both server and web and remove the server.xml default from here -- only providing a value via helm charts if an override is provided.

pega-embate:
> Since the base image is specific to a particular platform version it might make sense to do the same for both server and web and remove the server.xml default from here -- only providing a value via helm charts if an override is provided.

I think it makes sense to remove the default server.xml from here

APegaDavis:
> > Since the base image is specific to a particular platform version it might make sense to do the same for both server and web and remove the server.xml default from here -- only providing a value via helm charts if an override is provided.
> 
> I think it makes sense to remove the default server.xml from here

Default server.xml removed and associated test has been updated.

pega-embate (approved):

misterdorito (approved):

NekhilKotha:
@APegaDavis  Can we add documentation for newly injectable files. https://github.com/pegasystems/pega-helm-charts/tree/master/charts/pega#pega-configuration-files

NekhilKotha (changes requested):
Can we add documentation for newly injectable files. https://github.com/pegasystems/pega-helm-charts/tree/master/charts/pega#pega-configuration-files

NekhilKotha (approved):

taz-pega-work (changes requested):
<!-- Note:  A "Server" is not itself a "Container", so you may not
     define subcomponents such as "Valves" at this level.
     Documentation at /docs/config/server.html

While most of the comments appropriately describe what you edit in the yaml, this one doesn't say what the port defines here or what shutdown is.

APegaDavis:
With latest changes (and reflected in the updated PR description):
* Allow overrides of server.xml and web.xml via values.yaml
* Provide default server.xml in config/deploy directory
* No default provided for web.xml (it is Pega Platform release-specific)
* If a web.xml is added to config/deploy directory, it will be used as the default during template rendering
* If no web.xml exists in config/deploy and no web.xml override is provided, it will not be included

taz-pega-work (approved):
documentation content looks good.

PR #313, opened 2021-06-21
Updated Pega Installer Job deletion hook policy

RyanStan:
After successful execution of the InfinityInstall job, the job and pods stay around until they are either deleted or the namespace is cleaned up. This was originally done because we wanted to ensure we had access to logs, but in Pega Cloud we're now sending those logs to CloudWatch.

For Pega Cloud only, updated hook deletion policy to include hook-succeeded.  These changes have no functional impact for client managed cloud.


RyanStan:
Have to make change and reopen, made syntax error in the case where cleanAfterInstall is not specified

MadhuriArugula (approved):

pega-roska (approved):

khick77 (approved):

pega-talba (approved):

PR #318, opened 2021-06-28
Update instructions for manually running terratest tests

RyanStan:
As of Go v1.16, the environment variable GO111MODULE is set to on by default.  Since the tests rely on GOPATH instead of modules, you have to set GO111MODULE to off.

I also made it explicit that to run the tests via Intellij, you have to have the Ultimate edition.


RyanStan:
Need to clean the commit history, will reopen once it's more readable

RyanStan:
Had to make new pull request: https://github.com/pegasystems/pega-helm-charts/pull/319

PR #365, opened 2021-12-16
Add option to set ephemeral storage to installer job

s4rd1nh4:
Add option to set ephemeral storage to installer job to avoid error below:

/opt/pega/kit/scripts/setupDatabase.xml:904: Only : 5860936 KB of free space found in '/opt/pega/temp' - Not enough for Temp Directory install (~ 6400000 KB)

ref.: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage


s4rd1nh4:
@dcasavant

arvasrikanth (approved):

PR #371, opened 2022-01-07
Support custom sidecar containers in deployment spec, plus test.

pega-talba:
Support custom sidecar containers within a pod, adjacent to the tomcat container.

Adopts the pull request https://github.com/pegasystems/pega-helm-charts/pull/239 created by @mathijs55

### DIFF
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -531,7 +531,9 @@ tier:
 
 ### Sidecar Containers
 
-If the pod needs to run with one or more [sidecar containers](https://kubernetes.io/docs/concepts/workloads/pods/#how-pods-manage-multiple-containers), you can specify custom 'sidecarContainers' for your deployment tier. Each sidecar container must include a complete container definition, including a name, image, resources, etc.
+Pega supports adding sidecar containers to manage requirements for your Pega application services that live outside of the primary tomcat container. This may include company policy requirements, utility images, networking containers, or other examples. For an overview of the versatility sidecar containers present, see [How Pods manage multiple containers](https://kubernetes.io/docs/concepts/workloads/pods/#how-pods-manage-multiple-containers).
+
+You can specify custom `sidecarContainers` for your deployment tiers in the Pega Helm chart as shown in the example below. Each sidecar container definition must be a complete container definition, including a name, image, and resources.
 
 Example:
 
--- a/charts/pega/README.md
+++ b/charts/pega/README.md
@@ -531,7 +531,7 @@ tier:
 
 ### Sidecar Containers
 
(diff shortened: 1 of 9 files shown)


APegaDavis (approved):

charts/pega/README.md
```diff
@@ -529,6 +529,21 @@ tier:
...
+
+Example:
+
+```yaml
+tier:
+  - name: my-tier
+    custom:
+      sidecarContainers:
+        - name: SIDECAR_NAME
+          image: SIDECAR_IMAGE_URL
```
zitikay, line 544: IDK if it's worth calling out in this README, since if you're using a sidecar container you probably know or have some guide for the required `spec`... but a container needs some `command` and the `resources` requests. From first part of comment and sake of keeping README concise, I totally understand if it's intentional to leave this as is

misterdorito (approved):

charts/pega/README.md
pega-talba: Updated with more indication that you do need more than simply the name and image. I don't think the command is always necessary, but resources should be there at least, and there's likely to be more. Using `...` to denote extra stuff, other examples here use that notation.

RyanStan: I vote for leaving them out since they're not required and might make things more confusing.  Although maybe we could have some sort of phrasing that indicates you could expand upon the name and image to include more fields describing the container.

RyanStan (approved):

charts/pega/README.md
```diff
@@ -529,6 +529,25 @@ tier:
...
 ```
 
+### Sidecar Containers
+
+If the pod needs to run with one or more [sidecar containers](https://kubernetes.io/docs/concepts/workloads/pods/#how-pods-manage-multiple-containers), you can specify custom 'sidecarContainers' for your deployment tier. Each sidecar container must include a complete container definition, including a name, image, resources, etc.
+
+Example:
+
+```yaml
+tier:
```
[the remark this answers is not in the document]
pega-talba: We're going to exclude it from the values.yaml itself because it would be a little messy to include it and it's not the typical expectation to have this.

charts/pega/README.md
```diff
@@ -529,6 +529,25 @@ tier:
       serviceAccountName: MY_SERVICE_ACCOUNT_NAME
 ```
 
+### Sidecar Containers
+
+If the pod needs to run with one or more [sidecar containers](https://kubernetes.io/docs/concepts/workloads/pods/#how-pods-manage-multiple-containers), you can specify custom 'sidecarContainers' for your deployment tier. Each sidecar container must include a complete container definition, including a name, image, resources, etc.
```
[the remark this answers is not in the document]
pega-talba: Updated with some refactored text based on these concerns.

taz-pega-work (approved):
thanks for updating the doc yo!

PR #373, opened 2022-01-10
1.22 changes were missing new requirement: `pathType`

zitikay:
`pathType` is added to all Ingress paths. With it's absence in extensions/v1beta1 version, it defaults to `pathType: ImplementationSpecific`, but is now required in networking.k8s.io/v1 version. To keep same behavior, all `pathType` 's are set to the default value we would otherwise be getting


pega-talba (approved):

zitikay:
[getIngresses.txt](https://github.com/pegasystems/pega-helm-charts/files/7841985/getIngresses.txt)
Interestingly enough, apparently the k8s cluster does beta - v1 translation, and this is bidirectional. In attached file, I am working with k8s cluster running v1.21.6. For `helm-test-betaingress` namespace, we can see the deprecation warnings, but also valid spec for both apiVersions of Ingress. For `helm-test-v1ingress` namespace, I have dropped version comparison to be `>=v1.18.0` which v1.21.6 satisfies giving us the v1 spec, but here also we can see valid spec for both apiVersions

RyanStan (approved):

PR #375, opened 2022-01-11
Prevent pega helm charts from failing when global.certificates are no…

misterdorito:
…t specified.


APegaDavis (approved):

pega-roska (approved):

zitikay (approved):

terratest/src/test/pega/data/values_with_customcerts.yaml
```diff
@@ -0,0 +1,51 @@
...
+        passivationTimeSec: 900
+      replicas: 1
+      deploymentStrategy:
+        rollingUpdate:
+          maxSurge: 1
+          maxUnavailable: 0
+        type: RollingUpdate
+      livenessProbe:
+        port: 8081
+      # Values for test - web
```
pega-talba, line 19: Could remove these pieces, but not technically necessary to remove.

terratest/src/test/pega/data/values_with_customcerts.yaml
```diff
@@ -0,0 +1,45 @@
...
+        passivationTimeSec: 900
+      replicas: 1
+      volumeClaimTemplate:
+        resources:
+          requests:
+            storage: 5Gi
+      livenessProbe:
+        port: 8081
+      # Values for test - stream
+      custom:
```
zitikay, line 45: why do we even end on this empty `custom:` if it doesn't have an associated value? In added test, and pega-helm-charts/terratest/src/test/pega/pega-tier-deployment_test.go I don't actually see any requirement/verification that global.tier.custom is set

zitikay: FWIW, looks like as of latest commit it's past the yamllint issues. So if all goes well with the build, no need to change this, mainly just curious why we have a `custom` for any of these deployment spec, since it's never checked or enforced as far as I can tell

pega-talba: There is no actual need for it - it can be removed. I think the origin was this test yaml being copied from the custom sidecar yaml, which included this custom block.

charts/pega/templates/_pega-deployment.tpl
```diff
@@ -187,8 +187,10 @@ spec:
         - name: {{ template "pegaVolumeCredentials" }}
           mountPath: "/opt/pega/secrets"
         #mount custom certificates
+{{ if .root.Values.global.certificates }}
```
zitikay, line 190: minor nitpick: "#mount custom certificates" comment should be within the condition. Currently it will render the comment regardless of if you have custom certs in your values.yaml

PR #379, opened 2022-01-24
add workflow file for github actions

smootherbug:
US - Switch helm chart repo from Travis to Github

use cases covered:
1. pr build:
	- run build job alone
2. commits on any branch
	- run the build job
	- run deploy job - pages
3. Tag build
	- run the build job
	- run deploy job - both to Github releases and pages
4. Concurrent builds, cancel the current build, and run the latest one

To see sample runs
https://github.com/smootherbug/pega-helm-charts/actions

### DIFF
--- a/.github/workflows/github-actions-build.yml
+++ b/.github/workflows/github-actions-build.yml
@@ -1,0 +1,155 @@
+name: GitHub Actions for pega helm chart
+
+env:
+  HELM_URL: https://get.helm.sh
+  HELM_TGZ: helm-v3.2.4-linux-amd64.tar.gz
+  YAMLLINT_VERSION: 1.15.0
+  GO_VERSION: 1.13.1
+
+
+on: 
+  push:
+    tags:
+      - '*'
+    branches-ignore:
+      - gh-pages
+  pull_request:
+    branches: [ master ]
+
+concurrency:
+  group:  ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  run-build-job:
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Install Helm
+        run: |
+          sudo rm /usr/local/bin/helm
+          wget -q ${{ env.HELM_URL }}/${{ env.HELM_TGZ }}
+          tar xzf ${{ env.HELM_TGZ }}
+          mv linux-amd64/helm /usr/local/bin/helm
+
+      - run: helm version
+      
+      - name: Install yamllint
+        run: |
+          sudo pip install yamllint=="${{ env.YAMLLINT_VERSION }}"
+      
+      - name: Install node
+        uses: actions/setup-node@v2
+        with:
+          node-version: lts/*
+      
+      - name: Install npm dependencies
+        run: |    
+          npm install -g --save remark-cli to-vfile remark-preset-lint-recommended remark-validate-links remark-lint-no-dead-urls remark-message-control remark-preset-lint-markdown-style-guide remark-lint
+      
+      - name: Install GO
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - run: export  PATH=$PATH:/usr/local/go/bin
+      
+      
+
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      
+      - name: run yamllint
+        run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/github-actions-build.yml
```diff
@@ -0,0 +1,155 @@
...
+    branches-ignore:
+      - gh-pages
+  pull_request:
+    branches: [ master ]
+
+concurrency:
+  group:  ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
```
dcasavant, line 23: I recommend splitting the lint steps from the Go test steps so they can run in parallel and not block on another by failures. You will need to name the jobs and use dependencies to ensure they all run, but we could then organize it like this: https://github.com/dcasavant/pega-helm-charts-incubator/pull/1/files#diff-4fe883216c20c1d7d34f343f489f9b5a4ab47a0e254226494e9a95a679e493e4R17

dcasavant: By the way - that PR was a work in progress and should not be treated as 'done' :)

dcasavant (approved):

.github/workflows/github-actions-build.yml
smootherbug: seperated the lint and the go test jobs, now these will run in parallel, and the deploy job will start sunning once all the others run successfully

.github/workflows/github-actions-build.yml
```diff
@@ -0,0 +1,168 @@
...
+          files: |
+            pega-kubernetes-example.tar.gz
+            pega-openshift-example.tar.gz
+            pega-azure-aks-example.tar.gz
+            pega-aws-eks-example.tar.gz
+            pega-google-gke-example.tar.gz
+            pega-pivotal-pks-example.tar.gz
+          repo_token: ${{ secrets.NEW_SECRET }}
+
+      - name: deploy on github pages
```
kishorv10, line 162: It appears, publish to gh-pages is happening for non tag builds as well. Even on travis too, gh-pages commit is happening for non tag builds, which seems to be un-intentional. 
@dcasavant  Can you please confirm?

.github/workflows/github-actions-build.yml
```diff
@@ -0,0 +1,168 @@
...
+          mkdir $GITHUB_WORKSPACE/terratest/bin
+          export GOPATH=$GITHUB_WORKSPACE/terratest
+          curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
+          export PATH=$PATH:$GITHUB_WORKSPACE/terratest/bin
+          cd terratest/src/test
+          dep ensure
+          # grep "FAIL" -A 8 - filter unnecessary logs from the final output but show failing tests
+          #  ; test ${PIPESTATUS[0]} -eq 0 - check if go test was finished success. Without it pipe return exit code from last command.
+          # go test test/pega | grep "FAIL" -A 8 ; test ${PIPESTATUS[0]} -eq 0
+          # go test test/addons | grep "FAIL" -A 8 ; test ${PIPESTATUS[0]} -eq 0
```
kishorv10, line 94: Can above commented code be removed?

.github/workflows/github-actions-build.yml
```diff
@@ -0,0 +1,168 @@
...
+        with:
+          prerelease: false
+          files: |
+            pega-kubernetes-example.tar.gz
+            pega-openshift-example.tar.gz
+            pega-azure-aks-example.tar.gz
+            pega-aws-eks-example.tar.gz
+            pega-google-gke-example.tar.gz
+            pega-pivotal-pks-example.tar.gz
+          repo_token: ${{ secrets.NEW_SECRET }}
```
kishorv10, line 160: Rename the secret to reflect its usage.

.github/workflows/github-actions-build.yml
smootherbug: done

.github/workflows/github-actions-build.yml
smootherbug: done

.github/workflows/github-actions-build.yml
dcasavant: @kishorv10 I agree, it seems unintentional. We should probably change to tagged builds only for publishing to gh-pages.

smootherbug: made the changes to push to gh-pages only for tagged builds

kishorv10 (approved):

Issue #376, opened 2022-01-13
addons deployment (1.9.1) fails on Kubernetes version 1.22


rbogendoerfer:
When trying to deploy traefik through the addons deployment (chart version 1.9.1) the deployment fails on Kubernetes 1.22:
helm -n pegaaddons install pegaaddons pega/addons --values 1.9.1/addons.yaml
Error: INSTALLATION FAILED: failed to install CRD crds/crds.yaml: unable to recognize "": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"

Manally fixing the deprecated API version from "v1beta1" to "v1" in the files charts/aws-load-balancer-controller/crds/crds.yaml and charts/ingress-azure/crds/azureingressprohibitedtarget.yaml does not fully fix the problem. Then a new error occurs:
 helm -n pegaaddons install pegaaddons ./addons-1.9.1 --values 1.9.1/addons.yaml
Error: INSTALLATION FAILED: failed to install CRD crds/azureingressprohibitedtarget.yaml: CustomResourceDefinition.apiextensions.k8s.io "azureingressprohibitedtargets.appgw.ingress.k8s.io" is invalid: [spec.versions: Invalid value: []apiextensions.CustomResourceDefinitionVersion(nil): must have exactly one version marked as storage version, status.storedVersions: Invalid value: []string(nil): must have at least one stored version, metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111]

So beside the deprecated API versions also the spec for the azure ingress needs fixing.

axelgMS:
Can you try using 1.5.0-rc1 of the AGIC Addon for k8s 1.22.0 to support ingress v1 API ?

_helm install ingress-azure -f helm-config.yaml  application-gateway-kubernetes-ingress/ingress-azure --version 1.5.0-rc1_

RyanStan:
Merged the pr that was mentioned above so this should be fixed now.  Once we merge a pr for issue #22 we'll cut a release and bump the major version of the charts.

pega-roska:
This should now be fixed in 2.0.0

PR #380, opened 2022-01-24
Update addons chart dependencies for Kubernetes 1.22

RyanStan:
ISSUE #376 was recently created because some of our dependencies in the addons chart were using deprecated API versions that were not in Kubernetes 1.22.  This was causing a helm install of the chart to fail.

I'm not certain we want to merge this pr yet but I though it would be a good idea to open it.   I'm hesitant because the [ingress-azure](https://github.com/Azure/application-gateway-kubernetes-ingress) chart only added K8s 1.22 support in their most recent release candidate, 1.5.0-rc1.  [There's an issue open on their project](https://github.com/Azure/application-gateway-kubernetes-ingress/issues/1317) about releasing 1.5.0 but so far it hasn't had any positive feedback.  Apparently the GA release for AKS on K8s 1.22 uses the release candidate as well.

### DIFF
--- a/charts/addons/README.md
+++ b/charts/addons/README.md
@@ -19,19 +19,19 @@ Microsoft Azure Kubernetes Service (AKS)  | Application Gateway Ingress Controll
 
 ### Traefik
 
-Deploying Pega Platform with more than one Pod typically requires a load balancer to ensure that traffic is routed equally.  Some IaaS and PaaS providers supply a load balancer and some do not. If a native load balancer is not provided and configured, or the load balancer does not support cookie based session affinity, Traefik may be used instead.  If you do not wish to deploy Traefik, set `traefik.enabled` to `false` in the addons values.yaml configuration. For more configuration options available for Traefik, see the [Traefik Helm chart](https://github.com/helm/charts/blob/master/stable/traefik/values.yaml).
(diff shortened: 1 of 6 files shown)


zitikay:
> Apparently the GA release for AKS on K8s 1.22 uses the release candidate as well.


given this, I'm much more open to taking rc release

RyanStan:
new release of ingress-azure came out so this should be good to go now

taz-pega-work (approved):
looks good from a documentation stand point ;-)

zitikay (approved):
Assuming successful test runs + local 1.22 testing, looks good to me

zitikay (commented):
(tried removing my review)
We should consider that these changes aren't backwards compatible. ie using same values.yaml for new pega-helm-charts release will inevitably fail. They'd have to update their values.yaml with the same changes in charts/addons/values.yaml. Whether we conditionally choose the addons version based on k8s version, or bump the major pega-helm-charts version when these changes are released signifying (semver) non-backwards compatible changes, not totally sure

zitikay (approved):
LGTM, as discussed earlier this should probably be a major version increase as values.yaml for addons will have to be updated, but if this all works on pre 1.22 and 1.22+ then I don't see anything preventing us from releasing it (with major bump)

APegaDavis (approved):

Issue #22, opened 2019-09-05
Allow changing location of busybox container


voor:
If you do not have access to Docker Hub (or it is blocked), you should allow chart users to change the location of the busybox container.

cdancy:
@voor that's not a bad idea. As it stands now you could certainly do this very thing and just amend the template files in the necessary locations but I could certainly see it being made easier if we had the config up-front and center to twiddle. Care to send in a PR?

@dcasavant thoughts?

dcasavant:
@voor Added to the backlog.

PegaBasis:
Hi @dcasavant  What is the latest status of this issue? I would also like to be able to change the busybox image for the pega and pegasearch deployment.

rbogendoerfer:
This issue is causing major issues in client environments, where the K8s cluster does not have direct Internet access. To address this issue the helm charts need to be pulled and modified locally.

As a suggested solution the busybox containers should use the same repository and secret as the Pega images, if those are specified in the pega.yaml file (docker.registry section).

PR #392, opened 2022-02-28
Configurable busybox and dcasavant/k8s-wait-for image locations

RyanStan:
Opening up a draft pull request for issue #22 to make the location of the busybox and dcasavant/k8s-wait-for images configurable.


terratest/src/test/pega/utilities.go
```diff
@@ -64,10 +64,6 @@ func VerifyInitContainerData(t *testing.T, containers []k8score.Container, optio
 			require.Equal(t, "cassandra:3.11.3", container.Image)
 			//The cassandra svc name below is derived from helm release name and not .Values.global.deploymentName like search svc
 			require.Equal(t, []string{"sh", "-c", "until cqlsh -u \"dnode_ext\" -p \"dnode_ext\" -e \"describe cluster\" pega-cassandra 9042 ; do echo Waiting for cassandra to become live...; sleep 10; done;"}, container.Command)
-		} else if name == "wait-for-cassandra" {
```
RyanStan, line 67: I just removed this block since I noticed it was a duplicate of the one above it

charts/backingservices/values.yaml
```diff
@@ -29,43 +40,48 @@ srs:
...
+    # To configure either authentication method, when the elasticsearch domain requires an open internet connection, uncomment to set this parameter to "true".
+    # requireInternetAccess: true
 
 # This section specifies the configuration for deploying an internal elasticsearch cluster for use with SRS.
 # The configuration for rest of the values defined under 'elasticsearch' are to define the elasticsearch cluster
 # based on helm charts defined at https://github.com/elastic/helm-charts/tree/master/elasticsearch and may be modified
 # as per runtime and storage requirements.
 elasticsearch:
   imageTag: 7.10.2
+
```
RyanStan, line 66: Added a bunch of spacing just to make things look better

charts/backingservices/README.md
```diff
@@ -38,11 +38,14 @@ Note: Pega does **not** actively update the elasticsearch dependency in `require
 
 Configuration                       | Usage
 ---                                 | ---
-`enabled`                           | Enable the Search and Reporting Service deployment as a backing service.
-`deploymentName`                    | Specify the name of your SRS cluster. Your deployment creates resources prefixed with this string. This is also the service name for the SRS.
-`srsRuntime`                        | Use this section to define specific resource configuration options like image, replica count, cpu and memory resource settings in the SRS.
+`srs.enabled`                           | Enable the Search and Reporting Service deployment as a backing service.
```
RyanStan, line 41: What do others think?  Is it a good idea to add these prefixes?  Personally I think it is.  Since the users of this are probably aware of how Helm Charts work, it makes things clearer so that users know the full path that these keys should be at (and don't misplace them elsewhere in the values file)

khick77 (approved):

charts/backingservices/values.yaml
misterdorito: Careful -- do not incur the wrath of the yaml linter.

misterdorito (approved):

charts/backingservices/README.md
RyanStan: Yep, in the actual values file in the backservices chart they're under srs.  I just wasn't sure if we were trying to make the readme easy to understand by not including the srs bit.  This is probably a question for you Taz @taz-pega-work :)

zitikay: So they're already under the `srs` block, so just a doc change to give more accurate json coordinates?

zitikay: So they're already under the `srs` block, so just a doc change to give more accurate json coordinates?

zitikay (approved):
Outstanding doc question for Taz, but looks good to me (both code + doc changes)

taz-pega-work (changes requested):
while these values are ok in the table, it's not consistent with any of the tables that I see in the pega readme file...in general the content in the readme and the descriptions should use the yaml parameter format, but the value column in the tables don't really need this. I think you either update ALL of the tables or you leave these set of values as is and describe the parameters in the descriptions appropriate, which I think is more clear.

charts/backingservices/README.md
RyanStan: Yep that was the intention but I'm going to change it back

RyanStan:
@taz-pega-work, I updated the README again and removed the `srs` prefix.  Do you think the description for `busybox` looks good?  I tried to do something similar to what you sent for the elasticsearch description

charts/backingservices/README.md
```diff
@@ -43,7 +43,7 @@ Configuration                       | Usage
...
 `elasticsearch`                     | Define the elasticsearch cluster configurations. The [Elasticsearch](https://github.com/helm/charts/tree/master/stable/elasticsearch/values.yaml) chart defines the values for elasticsearch provisioning in the cluster.
 `srsStorage.provisionInternalESCluster` | <ul><li>Enable this parameter to provision an internally managed and secured Elasticsearch cluster to be used with the SRS (Requires you to run `$ make es-prerequisite NAMESPACE=<NAMESPACE_USED_FOR_DEPLOYMENT>`).</li><li>Disable this parameter to use your own Elasticsearch service with the SRS.</li></ul>
-`busybox` | When provisioning an internally managed Elasticsearch cluster, you can customize the location and pull policy of the BusyBox image used during the deployment process by specifying `srs.busybox.image` and `srs.busybox.imagePullPolicy`.
+`busybox` | When provisioning an internally managed Elasticsearch cluster, you can customize the location and pull policy of the BusyBox image used during the deployment process by specifying `busybox.image` and `busybox.imagePullPolicy`.
```
taz-pega-work, line 46: lil tweak to the description but looks good>>
To provision an internally managed Elasticsearch cluster, customize both the location and pull policy of the BusyBox Docker image used during the deployment process by specifying `busybox.image` and `busybox.imagePullPolicy`.

@RyanStan - what does location mean here...is this the repository in which the client pushes the image? If it's "image" should this be an image name?? Location here isn't very obvious to me.

RyanStan: It's a bit odd because that image field specifies both the location **and** the name of the image that's being pulled.  Since we're providing it mostly to let users configure the location of the image, I just decided to say location

charts/pega/README.md
```diff
@@ -137,6 +137,24 @@ docker:
     imagePullPolicy: "Always"
 ```
 
+## Customize Utility Image Locations
+The Pega chart, installer chart, and pegasearch charts use the busybox and dcasavant/k8s-wait-for images for various utility tasks.  
+You can configure where and how these images are pulled by modifying `utilityImages.busybox` and `utilityImages.k8s_wait_for`.
```
taz-pega-work, line 142: can we try to take a higher level view for this??
## Pega-provided utilities image
Pega deployment charts include two images, `busybox` and `dcasavant/k8s-wait-for` Dockers images, that you can add to your deployment to complete various tasks in your clusters. these utilities include:
>>>can you give us a list or state a business value??<<<
 To add them to your deployment, you configure the image name and images pull policy by modifying `utilityImages.busybox` and `utilityImages.k8s_wait_for` parameters as described below.

taz-pega-work: If you are including this in the SRS readme, then I think that readme needs the same introduction or put the main introduction in the main readme so you don't have to repeat content>>https://github.com/pegasystems/pega-helm-charts/blob/master/README.md and then link appropriately to the readmes from the main readme.

RyanStan: Since we're not "adding" them to the deployment, and they're required, I might try another stab at it.  The business value they're adding is being able to correctly deploy the charts, essentially.

## Customize Utility Images Locations
The Pega and Installer charts require the `busybox` and `dcasavant/k8s-wait-for` images to deploy successfully.  Since some clients may not want to use the default DockerHub registry to pull these images from, we let customers configure where these images are pulled from by exposing `utilityImages.busybox` and `utilityImages.k8s_wait_for`.  These fields must have a value.

RyanStan: I'm between section headings.  
- Busybox and k8s-wait-for utility images
- Utility images
- Pega-provided utility images

taz-pega-work: how about>>

Deploying with Pega-provided Busybox and k8s-wait-for utility images
To deploy Pega Platform, the Pega helm  chart requires the use of the busybox and dcasavant/k8s-wait-for images. For clients who want to pull these images from non-default DockerHub registry they must push and tag these images to a private repository, and then pull these images by specifying `busybox.image` and `busybox.imagePullPolicy` values as described below.

RyanStan: Here's the above with a few slight changes to make it more technically accurate.  The one thing I wasn't 100% on was the end of the last line.  They have to modify `k8s-wait-for.image` and `k8s-wait-for.imagePullPolicy` as well as `busybox.image` and `busybox.imagePullPolicy`, so to make it more succinct is why I was kept it at `busybox` and `k8s-wait-for`

## Deploying with Pega-provided busybox and k8s-wait-for utility images
To deploy Pega Platform, the Pega helm chart requires the use of the busybox and k8s-wait-for images.  For clients who want to pull these images from a registry other than Docker Hub, they must tag and push these images to another registry, and then pull these images by specifying `busybox` and `k8s-wait-for` values as described below.

taz-pega-work: looks good! When you have the pega and SRS charts updated, I can do a final review there! 

Thanks for the excellent collaboration!

RyanStan: Ok changes have been made!  Thanks for the help Taz!

taz-pega-work: sure thing!

charts/backingservices/README.md
```diff
@@ -34,6 +34,10 @@ Note: Pega does **not** actively update the elasticsearch dependency in `require
 * Update the elasticsearch `version` value, or
 * Disable deploying elasticsearch by setting `srs.srsStorage.provisionInternalESCluster` to false.
 
+## Deploying SRS with Pega-provided busybox images
+The busybox image is required to deploy SRS. For clients who want to pull this image from a registry other than Docker Hub, they must tag and push their image to the registry, and then pull it by specifying `busybox.image` and `busybox.imagePullPolicy`.
```
taz-pega-work: can you update the statement here as well so there's 1-to-1 consistency?? otherwise it looks great! Second sentence is perfect.>>>
The busybox image is required to deploy SRS.
to>
To deploy Pega Platform, the SRS helm chart requires the use of the busybox image.

RyanStan: The reason I didn't initially use the same "To deploy Pega Platform..." part is because the SRS chart is independent of the Pega Platform.  So how about

"To deploy the Search and Reporting Service as an independent module, the SRS helm chart requires the use of the busybox image."

RyanStan: Or maybe this is better?

"To deploy Pega Platform with the SRS backing service, the SRS helm chart requires the use of the busybox image."

taz-pega-work: sure this is good! thanks!

RyanStan: Done! :)

taz-pega-work (approved):

taz-pega-work (approved):

RyanStan:
Still need one approval from @pegasystems/searchservice

Anand-Kotamsetti (approved):
LGTM.

PR #396, opened 2022-03-07
Increase the failure threshold from 20 to 30 to provide more time for…

misterdorito:
… a node to start.


RyanStan (approved):

terratest/src/test/pega/pega-tier-deployment_test.go
```diff
@@ -187,7 +187,7 @@ func VerifyDeployment(t *testing.T, pod *k8score.PodSpec, expectedSpec pegaDeplo
 	//require.Equal(t, int32(10), pod.Containers[0].StartupProbe.TimeoutSeconds)
```
RyanStan, line 187: Should these still be commented out?

misterdorito: The test code is commented out due to an issue of our test infra that doesn't really allow us to run tests against different k8s server versions.

Issue #388, opened 2022-02-11
Convert terratest to use Go modules


RyanStan:
As we support multiple versions of Kubernetes on client-managed cloud (["Supports all vendor supported versions as detailed on the vendor site"](https://docs.pega.com/client-managed-cloud/83/meeting-requirements-and-prerequisites)), we have to update packages like k8s.io/api to account for deprecated apis.  

Working with specific package versions using `dep` is not a great experience.  Trying to upgrade the `k8s.io/api` constraint in Gopkg.toml is creating all kinds of dependency issues for me.  Go Modules would make life much easier and would let us use multiple versions of the same package within the project.  

Is there any reason why we haven't switched to Go modules yet?  If not, I'd be happy to open a pull request with the necessary changes.

One other thing to add.  This will make debugging much easier since `gopls`, the Go language server, requires Go Modules to be enabled

PR #399, opened 2022-03-14
Convert terratest to use Go modules

RyanStan:
As we support multiple versions of Kubernetes on client-managed cloud (["Supports all vendor supported versions as detailed on the vendor site"](https://docs.pega.com/client-managed-cloud/83/meeting-requirements-and-prerequisites)), we have to update packages like k8s.io/api to account for deprecated Kubernetes apis.

Working with specific package versions using dep is not a great experience. Trying to upgrade the k8s.io/api constraint in Gopkg.toml creates all kinds of dependency issues.  Go Modules would make life much easier and would let us use multiple versions of the same package within the project (i.e. via replace).

This will also make debugging much easier since gopls, the Go language server, requires Go Modules to be enabled.  Plus running tests will be easier since we won't have to mess with environment variables anymore.

Issue #388


APegaDavis (approved):

terratest/run-tests-from-cmdline.md
```diff
@@ -1,19 +1,11 @@
 # Running tests from the command line
 
-The test can also be run from the command line which allows you to avoid complying with the IntelliJ plugin license requirements.
+The tests can be run from the command line which allows you to avoid complying with the IntelliJ plugin license requirements.
```
taz-pega-work: rather than sounding naughty, how about this?
Run the tests using the command line to avoid using the IntelliJ plugin.

terratest/run-tests-from-cmdline.md
```diff
@@ -1,19 +1,9 @@
...
-- You have installed dep
-- You have disabled go modules by setting the environment variable `GO111MODULE="off"` (v1.16 or later)
-
-## Setup
-Add {path to repo}/terratest to the GOPATH environment variable on your local system.
-
-Then:
-- cd to {path to repo}/terratest/src/test
-- run `dep ensure` -- this may take a little while and log warnings.
+- You have enabled go modules by setting the environment variable `GO111MODULE="on"`
```
taz-pega-work, line 6: You enabled go modules by setting the environment variable `GO111MODULE="on"`

taz-pega-work (approved):

PR #404, opened 2022-03-17
add note about CPU support

pega-roska:


taz-pega-work (approved):
V sht&scct. ACE! thanks for checking! Looks great!

pega-talba (approved):

PR #418, opened 2022-05-06
PegaRULESLongRW does not take password

zitikay:
Main use case for PegaRULESLongRW datasource is for long running CDH queries. Verified with Martin Winkler - CDH owner - that the socketTimeout is the only reason they don't use main PegaRULES connection. This user for this connection never needed to be stored in a different env var. Keeping env var value if it exists for backwards compatibility purposes.


RyanStan:
Do RW and normal user have the same passwords?

RyanStan (approved):

zitikay:
This PegaRULESLongRW datasource is very clearly something used for Pega CloudK and not actually used in client-managed-cloud, based on the fact that it never even took a password which assumes a jdbc driver which can reach out to secrets manager directly. From [postgres jdbc driver doc](https://jdbc.postgresql.org/documentation/head/connect.html): "If a property is specified both in URL and in Properties object, the value from Properties object is ignored."
So by putting the LongRW connection properties directly in the URL, this will take precedence over the same properties set via connectionProperties

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -21,20 +21,20 @@
     minEvictableIdleTimeMillis="60000"
     />
 
-  {{ if and .Env.DB_RW_USERNAME .Env.JDBC_RW_URL }}
+  {{ if or .Env.SET_RW .Env.JDBC_RW_URL }}
```
zitikay, line 24: new env var: `SET_RW` which if set and no `JDBC_RW_URL` set will take the url for main PegaRULES datasource. The intended behavior of this datasource *was* to always be same as main PegaRULES just with longer timeouts. By still including JDBC_RW_URL we allow for backwards compatibility

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -21,20 +21,20 @@
...
     type="javax.sql.DataSource"
     driverClassName="{{ .Env.JDBC_CLASS }}"
-    url="{{ .Env.JDBC_RW_URL }}"
-    username="{{ .Env.DB_RW_USERNAME }}"
-    password=""
+    url="{{- if .Env.JDBC_RW_URL -}}{{ .Env.JDBC_RW_URL }}{{- else -}}{{ .Env.JDBC_URL }}{{- end -}}?socketTimeout=1200;statement_timeout=1200500;idle_in_transaction_session_timeout=1200500"
+    username="{{ .Env.SECRET_DB_USERNAME }}"
+    password="{{ .Env.SECRET_DB_PASSWORD }}"
     maxTotal="{{ .Env.JDBC_MAX_ACTIVE }}"
-    minIdle="{{ default .Env.JDBC_RO_MIN_IDLE .Env.JDBC_MIN_IDLE }}"
```
zitikay, line 33: not sure why we were using the RO settings and only defaulting to main settings if no RO were set...

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -21,20 +21,20 @@
...
+    minIdle="{{ .Env.JDBC_MIN_IDLE }}"
     maxIdle="{{ .Env.JDBC_MAX_IDLE }}"
     maxWaitMillis="{{ .Env.JDBC_MAX_WAIT }}"
-    initialSize="{{ default .Env.JDBC_RO_INITIAL_SIZE .Env.JDBC_INITIAL_SIZE }}"
-    connectionProperties="socketTimeout=1200;statement_timeout=1200500;idle_in_transaction_session_timeout=1200500;"
```
zitikay, line 37: these were moved to end of url above

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -21,20 +21,20 @@
...
-  {{ if and .Env.DB_RW_USERNAME .Env.JDBC_RW_URL }}
+  {{ if or .Env.SET_RW .Env.JDBC_RW_URL }}
     <Resource name="jdbc/PegaRULESLongRW"
     auth="Container"
     type="javax.sql.DataSource"
     driverClassName="{{ .Env.JDBC_CLASS }}"
-    url="{{ .Env.JDBC_RW_URL }}"
-    username="{{ .Env.DB_RW_USERNAME }}"
-    password=""
+    url="{{- if .Env.JDBC_RW_URL -}}{{ .Env.JDBC_RW_URL }}{{- else -}}{{ .Env.JDBC_URL }}{{- end -}}?socketTimeout=1200;statement_timeout=1200500;idle_in_transaction_session_timeout=1200500"
```
pega-roska, line 29: I can 100% guarantee you that this is not valid url syntax on one or more supported platforms, including postgres.  I have some doubts that the properties are all valid across supported platforms in general, but we probably should keep them in the connection properties attr so they at least parse properly.

zitikay: fair enough. Put them back in connection properties. Determined that the last property wins if there are multiple, so appending should be good enough

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -21,20 +21,20 @@
...
+    password="{{ .Env.SECRET_DB_PASSWORD }}"
     maxTotal="{{ .Env.JDBC_MAX_ACTIVE }}"
-    minIdle="{{ default .Env.JDBC_RO_MIN_IDLE .Env.JDBC_MIN_IDLE }}"
+    minIdle="{{ .Env.JDBC_MIN_IDLE }}"
     maxIdle="{{ .Env.JDBC_MAX_IDLE }}"
     maxWaitMillis="{{ .Env.JDBC_MAX_WAIT }}"
-    initialSize="{{ default .Env.JDBC_RO_INITIAL_SIZE .Env.JDBC_INITIAL_SIZE }}"
-    connectionProperties="socketTimeout=1200;statement_timeout=1200500;idle_in_transaction_session_timeout=1200500;"
+    initialSize="{{ .Env.JDBC_INITIAL_SIZE }}"
+    connectionProperties="{{ .Env.JDBC_CONNECTION_PROPERTIES }}socketTimeout=1200;statement_timeout=1200500;idle_in_transaction_session_timeout=1200500;"
```
pega-roska, line 37: Do we want to rely on there being a trailing semicolon in the user-supplied props?  Not sure that's a good idea.

zitikay: Fair enough. Would starting with a semicolon be a problem if there are no connection properties passed though? Guessing probably not, it would just count as a null/nil/empty connection property, just want to confirm

pega-roska (approved):

charts/pega/config/deploy/context.xml.tmpl
```diff
@@ -21,20 +21,20 @@
...
+    password="{{ .Env.SECRET_DB_PASSWORD }}"
     maxTotal="{{ .Env.JDBC_MAX_ACTIVE }}"
-    minIdle="{{ default .Env.JDBC_RO_MIN_IDLE .Env.JDBC_MIN_IDLE }}"
+    minIdle="{{ .Env.JDBC_MIN_IDLE }}"
     maxIdle="{{ .Env.JDBC_MAX_IDLE }}"
     maxWaitMillis="{{ .Env.JDBC_MAX_WAIT }}"
-    initialSize="{{ default .Env.JDBC_RO_INITIAL_SIZE .Env.JDBC_INITIAL_SIZE }}"
-    connectionProperties="socketTimeout=1200;statement_timeout=1200500;idle_in_transaction_session_timeout=1200500;"
+    initialSize="{{ .Env.JDBC_INITIAL_SIZE }}"
+    connectionProperties="{{ .Env.JDBC_CONNECTION_PROPERTIES }};socketTimeout=1200;statement_timeout=1200500;idle_in_transaction_session_timeout=1200500;"
```
zitikay, line 37: Verified that leading `;` and duplicate `;` do not cause issues. So whether there's no connection properties passed, or connection properties already end with `;`, beginning the hardcoded properties with `;` will not cause problems and is safety net if passed properties do *not* end in `;`

misterdorito (approved):

PR #437, opened 2022-07-21
BUG-737502 added infrastructure to give the user the option to change…

Evan1oconto:
… imagePullPolicy and made note of that in the README


charts/pega/README.md
```diff
@@ -782,7 +782,8 @@ Use the `installer` section  of the values file with the appropriate parameters
 Parameter   | Description   | Default value
 ---         | ---           | ---
 `image`   | Reference the `platform/installer` Docker image that you downloaded and pushed to your Docker registry that your deployment can access.  | `YOUR_INSTALLER_IMAGE:TAG`
-`adminPassword` | Specify a temporary, initial password to log into teh Pega application. This will need to be changed at first login. The adminPassword value cannot start with "@". | `"ADMIN_PASSWORD"`
+`imagePullPolicy` | Specify when to pull an image. | `IfNotPresent`
```
taz-pega-work, line 785: The intro paragraph to these parameters doesn't mention this and I don't believe clients have enough context to which "when" refers. Can you explain this context in the paragraph leading up to this table? you might have to also provide an example so it's clear what this new "image pull" policy is all about.

taz-pega-work (changes requested):
I think we need some explanation for the business value of an image pull policy. Please provide details in the readme outside of the table in the paragraph and maybe with an example.

APegaDavis (approved):

charts/pega/README.md
APegaDavis: Or we just don't document this feature :)

kishorv10 (approved):

pega-talba (approved):

khick77 (approved):

charts/pega/README.md
```diff
@@ -759,14 +759,18 @@ The Helm charts also support an automated install or upgrade with a Kubernetes J
 
 For installations of the Pega platform, you must specify the installer Docker image and an initial default password for the `administrator@pega.com` user.
 
+Along with this, you can specify the imagePullPolicy here, an element that defines when an image should be pulled. It is defaulted to `IfNotPresent`, meaning an image will be pulled if it is "not present". However, depending on how images are built, it may be worthwhile to change it to `Always`, as fixes may come out that aren't automatically applied to the image.
```
taz-pega-work: a few tweaks and still need more information about where the deployment pulls the images...please adjust>>
For installations of the Pega platform, you must specify the installer Docker image and an initial default password for the `administrator@pega.com` user.

Along with this, you can specify the imagePullPolicy here, an element that defines when an image should be pulled. It is defaulted to `IfNotPresent`, meaning an image will be pulled if it is "not present". However, depending on how images are built, it may be worthwhile to change it to `Always`, as fixes may come out that aren't automatically applied to the image.
change to>>
To install Pega Platform, you must specify the Pega installer Docker image that you downloaded and placed in your repository. Add an initial default password for the `administrator@pega.com` user.

You can specify a `imagePullPolicy`, which is a parameter that defines a use case for when a deployment pulls a Pega installer image from "where??????". By default, the policy is set to `IfNotPresent`; this specifies that the Pega installer image will be pulled if it is "not present" in your specified repository, defined using the `which-one????` parameter. Supported use cases are `IfNotPresent` and `Always`.
Change `imagePullPolicy` to `Always`, to configure your deployment to automatically pull any new Docker images that come out from "where???".

taz-pega-work (changes requested):
please update with more context as provided in my review comments

charts/pega/README.md
APegaDavis: @taz-pega-work `imagePullPolicy` is a standard kubernetes feature related to when to pull the docker image from the docker repository for a container included in a deployment or job. I don't think we want to document it ourselves.

taz-pega-work (approved):

Issue #442, opened 2022-07-27
New Traefik config not fully optional


APegaDavis:
**Describe the bug**
Recently a traefik template was introduced that requires [a new setting](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/templates/_pega-traefik-config.tpl#L4) in values provided during helm provisioning. This was handled in the shipped values files, but some configurations (like CloudK) do not use shipped values.

failure:
```
error while running command: exit status 1; Error: template: /charts/pega/templates/pega-traefik-config.yaml:6:4: executing "charts/pega/templates/pega-traefik-config.yaml" at <include "pegaTraefikConfigTemplate" $data>: error calling include: template: charts/pega/templates/_pega-traefik-config.tpl:4:12: executing "pegaTraefikConfigTemplate" at <.node.service.tls.traefik.enabled>: nil pointer evaluating interface {}.enabled
```
 

We could rewrite to be fully optional as follows:
```
{{- if (.node.service.tls.traefik) }}
{{- if (.node.service.tls.traefik.enabled) }}
```

**To Reproduce**
Deploy with values that do not include `.service.tls.traefik.enabled` for web-tier deployments

**Expected behavior**
Traefik should be ignored without changes

**Chart version**
2.2.0 or 2.3.0

PR #441, opened 2022-07-26
US-488811 - tomcat certs, keystore to use external secrets

smootherbug:

PR #443, opened 2022-07-27
Make traefik config fully optional

APegaDavis:
Address #442 and check that `.node.service.tls.traefik` exists before checking `.node.service.tls.traefik.enabled` so that users do not have to modify their values to include the latter.


RyanStan (approved):

APegaDavis:
changes picked up in #441

PR #482, opened 2022-11-17
Update README.md

ravisagar100:
Made changes as per the edit suggested


taz-pega-work (approved):
thank you for implementing my review suggestions. looks good!

PR #497, opened 2022-12-14
US-506187 - Enhance Cassandra client helm configuration options

baylp2022:
US-506187 - Enhance Cassandra client helm configuration options


charts/pega/values.yaml
```diff
@@ -347,6 +347,38 @@ dds:
   # username,password,trustStorePassword,keyStorePassword parameters in External Secrets Manager, and enter the external secret for the credentials
   # make sure the keys in the secret should be CASSANDRA_USERNAME, CASSANDRA_PASSWORD , CASSANDRA_TRUSTSTORE_PASSWORD , CASSANDRA_KEYSTORE_PASSWORD
   external_secret_name: ""
+  # Enable asynchronous processing of records in DDS Dataset save operation. Failures to store individual records will
+  # not interrupt Dataset save operations
```
taz-pega-work, line 351: add period>>operations.

charts/pega/values.yaml
```diff
@@ -347,6 +347,38 @@ dds:
...
+  asyncProcessingEnabled: false
+  # Specify a prefix to use when creating Pega-managed keyspaces in Cassandra
```
taz-pega-work, line 353: add period>>Cassandra.

charts/pega/values.yaml
```diff
@@ -347,6 +347,38 @@ dds:
...
+  # token from the token range to determine which Cassandra node to send the request. Before you can enable this policy,
+  # you must configure the token range partitioner.
+  extendedTokenAwarePolicy: false
+  # Enable a latency awareness policy, which collects the latencies of the queries for each Cassandra node and maintains
+  # a per-node latency score (an average).
+  latencyAwarePolicy: false
+  # Enable the use of a customized retry policy for your Pega Platform deployment. After enabling this policy in your
+  # deployment configuration, Cassandra queries that timeout will be retried. The number of retries may be configured
+  # using the dynamic system setting (DSS): dnode/cassandra_custom_retry_policy/retryCount. If not configured, queries
+  # will be retried once
```
taz-pega-work, line 365: add period>>once.

charts/pega/values.yaml
```diff
@@ -347,6 +347,38 @@ dds:
...
+  customRetryPolicy: false
+  # Enable the speculative execution policy for retrieving data from your Cassandra service. When enabled, the Pega
+  # Platform will send a query to multiple nodes in your Cassandra service and process the first query response. This
+  # provides lower perceived latencies at the client, but puts greater load on your Cassandra service.
+  speculativeExecutionPolicy: false
+  # Enable reporting of DDS SDK metrics to a Java Management Extension (JMX) format for use by your organization to
+  # monitor your Cassandra service. Setting this property `false` disables metrics being exposed through the JMX
+  # interface; disabling also limits the metrics being collected using the DDS landing page
```
taz-pega-work, line 373: add period>>page.

charts/pega/values.yaml
```diff
@@ -347,6 +347,38 @@ dds:
...
+  jmxMetricsEnabled: true
+  # Enable reporting of DDS SDK metrics to a Comma Separated Value (CSV) format for use by your organization to monitor
+  # your Cassandra service. If you enable this property, use the Pega Platform DSS:
+  # dnode/ddsclient/metrics/csv_directory to customize the filepath to which the deployment writes CSV files. By
+  # default, after you enable this property, CSV files will be written to the Pega Platform work directory
```
taz-pega-work, line 378: add period>>directory.

taz-pega-work (changes requested):
just a few missing periods are left now.

charts/pega/values.yaml
baylp2022: Fixed

charts/pega/values.yaml
baylp2022: Fixed

charts/pega/values.yaml
baylp2022: Fixed

charts/pega/values.yaml
baylp2022: Fixed

charts/pega/values.yaml
baylp2022: Fixed

taz-pega-work (approved):
looks great! Thanks for fixing all of these!

charts/pega/templates/_pega-deployment.tpl
```diff
@@ -163,6 +163,36 @@ spec:
         - name: REQUESTOR_PASSIVATION_TIMEOUT
           value: "{{ .node.requestor.passivationTimeSec }}"
 {{- end }}
+{{- if .node.dds }}
```
kishorv10, line 166: These settings are already added in `pega-environment-config.yaml` config map. Is is required to define them the deployment envs again? Per [this](https://github.com/pegasystems/pega-helm-charts/pull/497/files#diff-d20d9d381fee827b6a82e73797e5f07a86abfb7a208931ac0164a558178a90d3L186) all the entries in config maps are loaded to deployments envs.

baylp2022: This is intentional. Our requirement is to be able to specify these new configuration parameters system wide, but also on a per tier basis. I.e. at the same level as the node type. For example, we may want to enable speculative execution on decision hub nodes to ensure low latencies to Cassandra, but leave it disabled on data flow nodes to avoid overloading the Cassandra cluster during data flow runs.

kishorv10: I have two questions.
1. In the below snippet, params are referred from `.node.dds`. To this context .node is tier object. Is it that, to override at tier level these params are supposed to define in `.Values.tier[].dds.`? If my understanding is right, Why is there no mention of this configuration in 'Values.yaml`, do you want to keep them as a hidden feature?
2. Deployment object going to contain tier level overidden cassandra params along with global cassandra params coming from config map. There will be two entries for each param in the evn list. Have you verified which one is taking precedence?

baylp2022: The new configuration parameters are included for completeness. They are very rarely used in practice and would only be recommended based on detailed performance analysis of a system. Our desire is not to keep them hidden, but rather not promote their use where not required. The README.md file includes a reference example of how they may be specified at the tier level.

We have verified that the tier level settings take precedence over the global settings as per our requirement.

kishorv10: Thanks for the confirmation. This actually, brings in a use case of the tier-level override of environment settings. This feature is not yet available, I would prefer to provide it as generic feature rather limited to DDS. May be like `.Values.tier[].env,`
@dcasavant @petejo @pega-chikv  What are your views on this?

baylp2022: @kishorv10 Is it possible for you to approve this PR and if you wish to provide tier-level overrides as a generic feature we can move the DDS settings to be inline with any new approach at that time? Many thanks, Paul

kishorv10: Created enhancement request [Issue-499](https://github.com/pegasystems/pega-helm-charts/issues/499) to have Pega helm charts support tier-level environment settings override. This PR will be on hold until the enhancement is prioritized and made available.

kishorv10: @baylp2022  We realized that there is already feature (not exposed for general use) to pass custom environment variables at tier level, 
```
      livenessProbe:
        port: 8081

      custom:
        env: 
        - name: CENV_KEY1
          value: CValue1
        - name: CENV_KEY2
          value: CValue2
```
Can you please check if this helps your usecase?

kishorv10 (approved):

taz-pega-work:
with two approvals complete, merging content

PR #523, opened 2023-02-15
Fix typo of aws-iam-authenticator command

sotoiwa:
Fix typo. `ws-iam-authenticator` shoud be `aws-iam-authenticator`.


kingakowalska1 (approved):
Good catch, thanks for the fix!

taz-pega-work (approved):
thanks for addressing this

Issue #522, opened 2023-02-14
aws-load-balancer-controller pod is running, but log has error, and can not create ALB.


lxm1989:
**Describe the bug**
aws-load-balancer-controller pod is in running status, but log has error, and then can not create ALB. 
But I followed AWS manual(https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html) and no problem happened.
Here are the status and log.
<img width="525" alt="スクリーンショット 2023-02-14 135109" src="https://user-images.githubusercontent.com/18628014/218642543-23942703-b039-4285-a38f-1d9ad2089d8f.png">

<img width="960" alt="スクリーンショット 2023-02-14 134251" src="https://user-images.githubusercontent.com/18628014/218642481-cf39835c-711b-4567-af30-6432d6db54b7.png">

**To Reproduce**
1. curl -O https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.4.4/docs/install/iam_policy.json

2. aws iam create-policy \
    --policy-name AWSLoadBalancerControllerIAMPolicy \
    --policy-document file://iam_policy.json

3. eksctl create iamserviceaccount \
  --cluster=pega-86-demo \
  --namespace=pegaaddons \
  --name=aws-load-balancer-controller \
  --role-name AmazonEKSLoadBalancerControllerRole \
  --attach-policy-arn=arn:aws:iam::xxxxxxxxxxx:policy/AWSLoadBalancerControllerIAMPolicy \
  --approve

AWS manual(error did not happen):
4.1 helm install aws-load-balancer-controller eks/aws-load-balancer-controller \
  -n pegaaddons \
  --set clusterName=pega-86-demo \
  --set serviceAccount.create=false \
  --set serviceAccount.name=aws-load-balancer-controller 

Pega manual(error happened):
4.2 helm install addons pega/addons --namespace pegaaddons --values addons-eks.yaml

addons-eks.yaml:
<img width="907" alt="スクリーンショット 2023-02-14 144306" src="https://user-images.githubusercontent.com/18628014/218649796-2adca950-fb44-49a9-8016-4350b26b1454.png">

**Expected behavior**
Hope Pega can give a right manual.

**Chart version**
2.6.2
No customizations

**Server (if applicable, please complete the following information):**
 - Environment:Amazon EKS
 - Database: PostgreSQL

sotoiwa:
[...]

PR #526, opened 2023-02-16
fix: addon chart sample values for eks

sotoiwa:
Fixed incorrect parameter name for VPC ID in the sample values file for EKS of the add-on helm chart.
Related to #522


kingakowalska1 (commented):
In the official documentation the VPC ID parameter is referred to as "VpcID", for example here:
https://docs.aws.amazon.com/managedservices/latest/userguide/find-vpc.html

The examples I find in Stack Overflow, etc, also use "VpcID". 

Could you please explain why you want to change it?

Thanks in advance.

sotoiwa:
The official documentation for the aws-load-balancer-controller helm chart can be found here.
https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/helm/aws-load-balancer-controller/values.yaml#L144

The official documentation uses `vpcId`.

In fact, [the chart values file](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/addons/values.yaml#L55) in addon chart also uses `vpcId`. Only this example for EKS is wrong.

kingakowalska1 (approved):
Thank for the explanation and for the fix.

taz-pega-work (approved):
thanks for requesting the review

sotoiwa:
@taz-pega-work @kingakowalska1 Will this PR be merged? Do I need anything else to have it merged?

MadhuriArugula (approved):

PR #539, opened 2023-03-14
ISSUE-538 Ability to set the HTTP thread pool size

lukaszkarnasiewicz:


pega-talba (approved):

misterdorito (approved):

Heshma (approved):

APegaDavis (approved):

zitikay (approved):

APegaDavis:
@lukaszkarnasiewicz it might make sense to use a similar approach for the circuit breaker setting by adding a conditional block to the context.xml.tmpl in here and control it using an env variable.

APegaDavis:
You'll also need to update https://github.com/pegasystems/pega-helm-charts/blob/master/terratest/src/test/pega/data/expectedInstallDeployServer.xml.tmpl to fix the tests

khick77 (approved):

pega-talba:
If these settings should be used by CMC clients, then this PR should also include an update to https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/templates/pega-environment-config.yaml and https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/values.yaml (and probably also https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/values-large.yaml) to allow specifying this environment variable in CMC deployments. Otherwise, this would only be useful from a Pega Cloud perspective.

petejo:
Why aren't we also updating the pega-environment-config template, and the values.yaml?

lukaszkarnasiewicz:
> If these settings should be used by CMC clients, then this PR should also include an update to https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/templates/pega-environment-config.yaml and https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/values.yaml (and probably also https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/values-large.yaml) to allow specifying this environment variable in CMC deployments. Otherwise, this would only be useful from a Pega Cloud perspective.

Yes, I agree. This is just a first step and can/should be followed up to set more optimal defaults. But this is outside of the scope of our current effort - for now we're not ready to change the global default for the "web" tier. More discussion and performance tests are needed to reach a safe consensus. Argus team has done extensive tests for CDH inbound and we'll change the default maxThreads to 16 for the "dh" tier for any new CDH deployments (via cdhfull.json)

PR #551, opened 2023-04-07
BUG-793879 - Added K8s version support information

kingakowalska1:
Added a paragraph in the README about supported Kubernetes versions in Pega Helm Charts.


README.md
```diff
@@ -23,6 +23,8 @@ Because open-source Kubernetes depends on your specific configuration, Pega does
 
 The helm charts currently only support running on Kubernetes nodes with x86-64 CPUs.  ARM CPUs are currently unsupported.
 
+The helm charts support running on Kubernetes versions that are in Maintenance and Active Support at the time of the helm chart’s release. For more information, see [Kubernetes EOL policy](https://endoflife.date/kubernetes).
```
taz-pega-work, line 26: Based on that referenced article, I think it would be better to use the Kubernetes preferred terminology releases. As the bug states, it might be useful to add in details about how to get to the helm charts release page>>
The helm charts support running on Kubernetes releases that are in Maintenance and Active Support at the time of the helm chart’s release. To review the data of a Pega helm chart release, see [Releases](https://github.com/pegasystems/pega-helm-charts/releases); to review the latest matrix of Kubernetes release support, see [Kubernetes EOL policy](https://endoflife.date/kubernetes).

taz-pega-work (commented):
I think it would be good to make it very easy for a client to find the Pega Helm chart releases article on Github. Not all clients might understand how to find it. Also suggesting to stick with the  term release instead of version.

README.md
kingakowalska1: Great suggestions, thanks!

taz-pega-work (approved):

README.md
```diff
@@ -23,6 +23,8 @@ Because open-source Kubernetes depends on your specific configuration, Pega does
 
 The helm charts currently only support running on Kubernetes nodes with x86-64 CPUs.  ARM CPUs are currently unsupported.
```
petejo, line 24: I realize this was not changed, but it is not accurate.  It has nothing to do with the Helm Charts, Pega Infinity does not support running on ARM CPUs because it leverages libraries that are not compatible with ARM CPUs.   Once Pega Infinity is updated to support ARM CPUs, no changes are required to the Helm Charts.   This ARM CPU restriction information should be moved to the Platform Support Guide, and shouldn't be mentioned here, we cannot have multiple documentation locations to itemize hardware and software support.

kingakowalska1: I removed the paragraph from the README and will work with Susan Stewart to add it to the Platform Support Guide instead. 
Is the addition about K8s release information correct, or is there anything else you wish to change here? Thanks in advance.

petejo: Yes, the supported K8S versions statement is accurate.

kingakowalska1: Thank you for the confirmation. Could you please also approve the pull request so I can merge it?

misterdorito (approved):

PR #560, opened 2023-04-20
Derive the REST URL for ZDT upgrades.

misterdorito:
Derive the REST URL for ZDT upgrades so customers don't need to provide it.


kingakowalska1 (changes requested):
Hi, 

Could you please capitalize Pega Platform in the values.yaml files? 

Currently it's:
# Specify the username and password to access the pre-upgraded Pega platform via REST service to perform pre- and post-update actions during zero-downtime updates.

And should be:
# Specify the username and password to access the pre-upgraded Pega Platform via REST service to perform pre- and post-update actions during zero-downtime updates.

Thanks!
Kinga

alemax22:
I had a look at your changes, can you slightly change it and take into consideration the fact that the port used to expose the web services may not always be 80 or 443? In our deployments we are currently using the port 8080 also for the http traffic between the pods.
Thank you very much!

misterdorito:
@alemax22 thanks for taking a look!

My changes leverage the DNS name of the k8s service (typically listening on port 80 or 443, but that can be customized).

Typically the pods associated with a service listen on 8080.

In order to come up with the rest URL for the preupgrade pod to make the REST call to of the existing web pods directly, you would have to know the pod's DNS name or its IP address -- either of these things you would need to determine and both of them are ephemeral.  Maybe be it is unlikely the pod that you pick to connect to will go away prior to performing the REST call, but it is something within the realm of possibility.

The service name on the other hand isn't going to change (even if pods come and go).

The changes in this PR account for changes to the default k8s service ports (see the unit tests).

Given this explanation, do you still have concerns?

alemax22:
> @alemax22 thanks for taking a look!
> 
> My changes leverage the DNS name of the k8s service (typically listening on port 80 or 443, but that can be customized).
> 
> Typically the pods associated with a service listen on 8080.
> 
> In order to come up with the rest URL for the preupgrade pod to make the REST call to of the existing web pods directly, you would have to know the pod's DNS name or its IP address -- either of these things you would need to determine and both of them are ephemeral. Maybe be it is unlikely the pod that you pick to connect to will go away prior to performing the REST call, but it is something within the realm of possibility.
> 
> The service name on the other hand isn't going to change (even if pods come and go).
> 
> The changes in this PR account for changes to the default k8s service ports (see the unit tests).
> 
> Given this explanation, do you still have concerns?

Now I got it right, I haven't seen the condition on the service port, which as you said, can be customized and we have customized by putting 8080.
Thank you very much!

APegaDavis (approved):

misterdorito:
> Hi,
> 
> Could you please capitalize Pega Platform in the values.yaml files?
> 
> Currently it's:
> 
> # Specify the username and password to access the pre-upgraded Pega platform via REST service to perform pre- and post-update actions during zero-downtime updates.
> And should be:
> 
> # Specify the username and password to access the pre-upgraded Pega Platform via REST service to perform pre- and post-update actions during zero-downtime updates.
> Thanks! Kinga

@kingakowalska1 changes made.

charts/pega/values.yaml
```diff
@@ -421,7 +421,7 @@ installer:
     # For postgres databases that you are upgrading from Pega Infinity version 8.4.0 and later
     # And for Oracle databases that you are upgrading from Pega Infinity version 8.4.3 and later.
     targetDataSchema: ""
-    # Specify the username and password to access the pre-upgraded Pega platform via REST service to perform pre- and post-update actions during zero-downtime updates.
+    # Specify the username and password to access the pre-upgraded Pega Platform via REST service to perform pre- and post-update actions during zero-downtime updates.
```
taz-pega-work, line 424: For postgres databases that you are updating from Pega Infinity version 8.4.0 and later.
And for Oracle databases that you are updating from Pega Infinity version 8.4.3 and later.

Specify the username and password that the REST service uses to access the pre-update Pega Platform deployment. The process requires this access to complete pre- and post-update actions during zero-downtime updates.

taz-pega-work (commented):
Sorry but had a few more tweaks for you to consider for consistency

charts/pega/charts/installer/templates/pega-upgrade-environment-config.yaml
```diff
@@ -36,9 +36,49 @@ data:
 {{- if .Values.upgrade.upgradeSteps }}
   UPGRADE_STEP: {{ .Values.upgrade.upgradeSteps }}
 {{- end }}
-{{- if .Values.upgrade.pegaRESTServerURL }}
-  PEGA_REST_SERVER_URL: {{ .Values.upgrade.pegaRESTServerURL }}
-{{- end }}
+{{- $depName := "pega" }}
```
Heshma, line 39: @misterdorito minor comment -- can we move this logic to helpers.tpl?

Heshma (approved):

charts/pega/charts/installer/templates/pega-upgrade-environment-config.yaml
misterdorito: Moved.

charts/pega/values.yaml
misterdorito: I agree with you that it is inconsistent.

On this entire page, there are:
- 99 references to some form of the word "upgrade"
- 17 references to some form of the word "update"

"upgrade" is is used inconjunction with zero-downtime and also referenced in the title.

I'm happy to make the changes you suggested, but the result will still be inconsistent.

kingakowalska1 (approved):
Thanks for making the changes!

taz-pega-work (commented):
looks good. thanks!

PR #561, opened 2023-04-28
EPIC-77287: Enabling External Kafka by default

anuragh27crony:
- As part Externalization of service for Infinity'23, Stream needs to default to external mode.
- Once an existing deployment once externalized, Infinity '23 will delete the embedded artifacts at the next startup.


kingakowalska1 (changes requested):
Hi Anurag, 

I would like to tweak the comment a little:

Instead of:

# Enabled by default. When enabled deployment will use external Kafka as per configuration and doesn't use the "Stream" node type. Disable this setting to use the "Stream" node type.

Could we use: 

# Enabled by default. When enabled, the deployment uses external Kafka as configured below and does not use the "Stream" node type at the tier level. Disable this setting to use the embedded "Stream" node type.

Thanks!
Kinga

kingakowalska1 (approved):
Thanks! Looks good.

kishorv10:
@anuragh27crony  How do these changes behave during upgrades? Would it cause downtime? After the upgrade, external kafka is replaced by embedded stream automatically?

kishorv10:
@anuragh27crony  Is this duplicate of [584](https://github.com/pegasystems/pega-helm-charts/pull/584)? If so can you please withdraw.

anuragh27crony:
> @anuragh27crony Is this duplicate of [584](https://github.com/pegasystems/pega-helm-charts/pull/584)? If so can you please withdraw.

Yes..i will be withdrawing it.

PR #563, opened 2023-05-04
BUG-752454: Updated values.yaml for hazelcast section

Pushpendrak18:
BUG-752454: Updated values.yaml for hazelcast section

### DIFF
--- a/charts/pega/values-large.yaml
+++ b/charts/pega/values-large.yaml
@@ -442,13 +442,14 @@ hazelcast:
 
   # No. of initial members to join
   replicas: 3
-  # UserName to be used in client-server Hazelcast model for authentication
+  # UserName in the client-server Hazelcast model authentication. This setting is exposed and not secure.
   username: ""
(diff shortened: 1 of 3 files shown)


kingakowalska1 (changes requested):
Hi,

The changes look good, I would just make a small suggestion:

Instead of:
# To avoid exposing username and password parameters, leave these parameters empty and configure
# these cluster settings using an External Secrets Manager using the following keys in the secret
# for username: HZ_CS_AUTH_USERNAME and for password: HZ_CS_AUTH_PASSWORD.

Could we write:

# To avoid exposing username and password parameters, leave these parameters empty and configure
# these cluster settings using an External Secrets Manager. Use the following keys in the secret: 
# HZ_CS_AUTH_USERNAME for username and HZ_CS_AUTH_PASSWORD for password.

What do you think?

Thanks,
Kinga

kingakowalska1 (approved):
Looks good! Thanks.

Saurabh-16 (approved):

Issue #503, opened 2023-01-04
SRS NetworkPolicy doesn't work with Openshift


maracle6:
**Describe the bug**
The backingservices chart installs a NetworkPolicy for SRS that includes egress rules.  These rules do not allow for DNS resolution on an Openshift 4.10 cluster with networkType: OVNKubernetes which prevents the SRS init container from completing.

**To Reproduce**
Use an Openshift 4.10 cluster with networkType: OVNKubernetes and provisionInternalESCluster: true.  The SRS pod should stall on init 0/1, due to wget in the wait-for-internal-es-cluster init container failing with a 'bad address' error when looking up the elasticsearch service DNS name.

**Expected behavior**
Init container successfully checks the elasticsearch endpoint and SRS starts.

**Chart version**
2.5.1

**Additional context**
In our specific use case, we were able to fix the issue by editing the network policy to allow egress to namespace openshift-dns on UDP port 5353.  In addition we had to change the matchLabels to use 'kubernetes.io/metadata.name' instead of just 'name'.  Currently we need to customize the charts or manually edit the network policy after every installation or upgrade of the chart.

```yaml
  egress:
    - ports:
        - protocol: TCP
          port: 9200
      to:
        - podSelector:
            matchLabels:
              app: elasticsearch-master
    - ports:
        - protocol: UDP
          port: 5353
      to:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: openshift-dns
```

We need a way to add this egress (support provider option for backingservices chart, or a new option to use openshift egress rules).  Or, it might also be useful to support replacing the default egress rule with one entered into the values.yaml.

arvasrikanth:
As changes are merged and part of latest release , closing this issue

PR #564, opened 2023-05-04
Fixed SRS network-policy issue for OpenShift

bhowd1:
- Fixed SRS network-policy issue for OpenShift (Issue #503)
- Added egress rule for openshift-dns incase of openshift cluster.


bhowd1:
Raised a different PR to address the issue.

Issue #565, opened 2023-05-04
Add securityContext to installer job template


zitikay:
Similar to how [pega deployment template](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/templates/_pega-deployment.tpl#L115-L132) has some entry points to securityContext, the installer job should have entry points for adding security context via values.yaml

This is so that we can address static security failures in our internal builds. Examples of failures due to lack of securityContext:
https://docs.bridgecrew.io/docs/bc_k8s_37
https://docs.bridgecrew.io/docs/bc_k8s_22
https://docs.bridgecrew.io/docs/bc_k8s_19
https://docs.bridgecrew.io/docs/bc_k8s_34
https://docs.bridgecrew.io/docs/bc_k8s_29
https://docs.bridgecrew.io/docs/bc_k8s_28

PR #566, opened 2023-05-04
Add security context entrypoint into installer job

zitikay:
#565 provide ability to inject securityContext block into installer job


APegaDavis (approved):

misterdorito (approved):

zitikay:
was able to set securityContext settings through pega chart values.yaml (under `installer: securityContext`)

PR #568, opened 2023-05-10
BUG-798792 - Using latest Helm charts during update

kingakowalska1:
Added a note recommending the use of the latest Helm version and Helm chart release when updating Pega Platform.


taz-pega-work (approved):
thanks for doing this!

PR #569, opened 2023-05-11
run helm lint as part of build

pega-roska:


pega-roska:
Leaving this as a draft until we get #562 in and actually fix all the existing lint issues so the check can pass

misterdorito (approved):

pega-talba (approved):

PR #575, opened 2023-06-05
added TLS support for openshift Route

pega-sagas1:


charts/pega/values-large.yaml
```diff
@@ -183,6 +183,10 @@ global:
           # For GKE Pre-shared Certificate, use ingress.gcp.kubernetes.io/pre-shared-cert: <pre-shared-certificate-name>
           # For GKE to use static IP for load balancer, use kubernetes.io/ingress.global-static-ip-name: <global-static-ip-name>
           ssl_annotation:
+          # for enabling TLS in Openshift Route as Reencrypt, add certificate, key and CaCertificate PEM or crt format.
```
taz-pega-work, line 186: for enabling TLS in Openshift Route as Reencrypt, add certificate, key and CaCertificate PEM or crt format.

can we update this wording to be less jargony? maybe not all clients would understand as written. As a start, I suggest>
For Openshift, enable TLS in Route do this [whatever Reencrypt specifically means in this context], and add your certificate, the corresponding key using the appropriate .pem or .crt format.

taz-pega-work (commented):
why is this change only available in values-large.yaml and not in the other two Pega

charts/pega/values-large.yaml
taz-pega-work: also wondering why this is only in large-values.yaml and not in the basic values.yaml file.

pega-sagas1:
hi @taz-pega-work, Please review.

charts/pega/values-large.yaml
```diff
@@ -183,7 +183,7 @@ global:
           # For GKE Pre-shared Certificate, use ingress.gcp.kubernetes.io/pre-shared-cert: <pre-shared-certificate-name>
           # For GKE to use static IP for load balancer, use kubernetes.io/ingress.global-static-ip-name: <global-static-ip-name>
           ssl_annotation:
-          # for enabling TLS in Openshift Route as Reencrypt, add certificate, key and CaCertificate PEM or crt format.
+          # For Openshift, to enable TLS, the termination policy would be reencrypt rather than edge, and we can add our certificate, the corresponding key using the appropriate .pem or .crt format and specify a CA certificate to validate the endpoint certificate, securing the connection from the browser to the router.
```
taz-pega-work: can we refine this??
For Openshift, Pega  deployments enable TLS to secure the connection from the browser to the router by creating the route using reencrypt termination policy. Add your certificate, the corresponding key using the appropriate .pem or .crt format and specify a CA certificate to validate the endpoint certificate.

taz-pega-work (changes requested):
Still need to improve sentence structure. Did I capture the gist of the requirements?

taz-pega-work (approved):
thanks for collaborating!

taz-pega-work (approved):
thanks for collaborating

MadhuriArugula (approved):

PR #610, opened 2023-08-08
US-512993 : Adding the capability to configure servicetype.

anuds20:
Added the capability to specify the service type. Also added a capability to restrict access to specific IP CIDR blocks in case of a LoadBalancer service type.

### DIFF
--- a/charts/pega/values.yaml
+++ b/charts/pega/values.yaml
@@ -174,7 +174,6 @@ global:
             serverName: ""
             # set insecureSkipVerify=true, if the certificate verification has to be skipped
             insecureSkipVerify: false
-        
 
       ingress:
         # For help configuring the ingress block including TLS, see the Helm chart documentation
--- a/charts/pega/values.yaml
+++ b/charts/pega/values.yaml
@@ -168,7 +168,7 @@ global:
         # Specify the CIDR ranges to restrict the service access to the given CIDR range.
         # Each new CIDR block should be added in a separate line.
         # Should be used only when serviceType is set to LoadBalancer.
-        # e.g 
+        # e.g
         # loadBalancerSourceRanges:
         #     - "123.123.123.0/24"
         #     - "128.128.128.64/32"
--- a/charts/pega/values.yaml
+++ b/charts/pega/values.yaml
@@ -138,6 +138,18 @@ global:
         httpEnabled: true
         port: 80
         targetPort: 8080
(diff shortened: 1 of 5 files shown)


charts/pega/values.yaml
```diff
@@ -162,12 +162,16 @@ global:
...
             insecureSkipVerify: false
-        serviceType: "LoadBalancer"
-        # Specify the CIDR range inside the qoutations to restrict the service access to the given CIDR range
-        # e.g "123.123.123.0/24"
-        loadBalancerSourceRanges:
-          - "103.38.88.0/24"
-          - "49.43.219.90/32"
+        # Use this parameter to deploy service of a specifictype. This is an optional value and should be used based on the usecase.
+        # serviceType: ""
+        # Should be used only when serviceType is set to LoadBalancer.
```
taz-pega-work, line 167: use spacing between words. If servicetype requires, LoadBalancer without a space, you have to call that out.

Is this the context? I'm guessing here...please add any missing details>
To specify a load balancing service, uncomment the previous line to enable its use and specify, `serviceType:` "LoadBalancer".

charts/pega/values.yaml
```diff
@@ -162,12 +162,16 @@ global:
...
+        # Specify the CIDR ranges to restrict the service access to the given CIDR range.
+        # Each new CIDR block should be added in a separate line.
```
taz-pega-work, line 169: Uncomment the following lines and add each CIDR block in a separate line using double quotes, replacing the example with your configuration requirements.

taz-pega-work (changes requested):
Please update the inline comments to be a little more precise and helpful to clients. I hope this helps.

charts/pega/values.yaml
anuds20: Made changes according to the comments.

taz-pega-work (approved):
looks much better. One more small change would be helpful if you can make the change on line 144>>
For example if you want to deploy a service of type LoadBalancer, uncomment the following line and specify serviceType: "LoadBalancer"
        # serviceType: ""

charts/pega/values.yaml
anuds20: Made changes according to the comment.

anuds20:
> looks much better. One more small change would be helpful if you can make the change on line 144>> For example if you want to deploy a service of type LoadBalancer, uncomment the following line and specify serviceType: "LoadBalancer" # serviceType: ""

Made changes according to the comment

MadhuriArugula (approved):

PR #614, opened 2023-08-14
🐛 fix(build): Added `external_secrets_name` condition for Cassandra auth

DylanAlloy:
When providing values to the username and password via templating, the `CASSANDRA_USERNAME` and `CASSANDRA_PASSWORD` environment variable instantiation portion of the the template does not check for `external_secrets_name`, e.g. the JDBC connection. This causes a crash of Pega docker images in a kubernetes environment if the engineer uses `external_secrets_name` in the DDS block of `values.yaml`.

Simply adding the same `if` logic from the JDBC portion of the code fixes this and allows the chart to finish templating `server.xml.tpl` and other required tomcat files given deployments which have pre-existing secrets, or ones being added to `./templates/`.

### DIFF
--- a/charts/pega/templates/_pega-credentials-secret.tpl
+++ b/charts/pega/templates/_pega-credentials-secret.tpl
@@ -35,9 +35,13 @@ data:
 
  {{ if (eq (include "performDeployment" .) "true") }}
   # Base64 encoded username for connecting to cassandra
+  {{ if .Values.dds.username -}}
   CASSANDRA_USERNAME: {{ .Values.dds.username | b64enc }}
+  {{- end }}
   # Base64 encoded password for connecting to cassandra
+  {{ if .Values.dds.password -}}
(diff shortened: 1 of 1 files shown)


jalsanishant:
any timeline when this will be addressed? thanks.

kishorv10:
@DylanAlloy @jalsanishant There has been refactoring on the secrets, the use case you are referring to should have been addressed. 
I kindly request that you retest the use case and provide us with the results when possible. Your feedback would be greatly appreciated.

MadhuriArugula:
@DylanAlloy @jalsanishant Have you got a chance to test the use case with latest charts? 
We shall proceed with the PR closure if we do not have the response from you in next 7 days.

DylanAlloy:
> @DylanAlloy @jalsanishant Have you got a chance to test the use case with latest charts? We shall proceed with the PR closure if we do not have the response from you in next 7 days.

all good. Thank you!

PR #625, opened 2023-08-31
Adding new base configuration files

pardhu-pokuri:
Adding base configuration files for tomcat web settings, java security override and catalina properties


charts/pega/config/deploy/tomcat-web.xml
Saurabh-16, line 1: Can we make this file candidate for xml lint check. Sometimes we unintentionally disturb the xml formatting resulting in xml parsing issue.
See reference here
https://github.com/pegasystems/pega-helm-charts/blob/master/validatexml.sh

charts/pega/config/deploy/catalina.properties
```diff
@@ -0,0 +1,219 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
Saurabh-16, line 1: I am assuming we have taken the source code from tomcat container we run during deployments?

saikarthik528: Correct.  We picked the default file

charts/pega/config/deploy/tomcat-web.xml
saikarthik528: Addressed

Saurabh-16 (approved):

kishorv10 (approved):

PR #634, opened 2023-09-22
Only set custom.jvm.args if CUSTOM_JVM_ARGS is set

APegaDavis:
The installer setting custom.jvm.args requires setting the max heap size if provided. Currently CUSTOM_JVM_ARGS is not set by default, so the installer is passed custom.jvm.args which bypasses the default max heap size in the installer.

Conditionalize the inclusion of custom.jvm.args bsaed on whether it is actually being configured.


pega-talba (approved):

khick77 (approved):

zitikay (approved):

kishorv10 (approved):

APegaDavis:
Closing in favor of https://github.com/pegasystems/pega-helm-charts/pull/639

Issue #581, opened 2023-06-14
Resource elasticsearch-master-pdb not found while deploying backingservices using helm repository and k8s 1.25


efiku:
**Describe the bug**

You are unable to deploy backingservices using the helm repository option (no local repository) because requirements.yaml in directory backingservices is not updated with ES version. 

**Error received:**

Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "elasticsearch-master-pdb" namespace: "" from "": no matches for kind "PodDisruptionBudget" in version "policy/v1beta1"
ensure CRDs are installed first

So, the backingservices charts depends on the ES chart version. 
Should be version: "7.17.3" but its still 7.10.2

**This is working fine for K8s  < 1.25. but giving a lot of warring about v1 beta apis**

**To Reproduce**
Steps to reproduce the behavior.

1. helm repo add pega https://pegasystems.github.io/pega-helm-charts
2. helm search repo pega , 3.1.0 everywhere
3. helm inspect values pega/backingservices > backing.yaml
4. helm install srs pega/backingservices -n srs -f backing.yaml

**Expected behavior**
Correct backingservices deployment for k8s 1.25 api

**Chart version**
Latest - using helm repository (no local repo)

**Server (if applicable, please complete the following information):**
 - OS: Ubuntu 20
 - Environment: Kubernetes 1.25
 - Database: PostgreSQL

**Additional context**
Suggested fix: changing the requirements.yaml es version to 7.17.3 in backingservices + also in the example values file.

**Workaround:** 
1. Clone repository to local, change above with correct 7.17.3 version of ES and update dependencies.
2. Change [requirements.yaml](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/backingservices/requirements.yaml#L8)   with 7.17.3 for example
3. Change [values.yaml](https://github.com/pegasystems/pega-helm-charts/blob/master/charts/backingservices/values.yaml#L82) to 7.17.3 
4. Update dependencies 
5. Redeploy backingservices from local repo using helm.

https://github.com/pegasystems/pega-helm-charts/blob/09393af4942456f24c133aae36713e2cce2862b0/charts/backingservices/requirements.yaml#L3-L5

rcomanne:
[...]

PR #647, opened 2023-10-11
US-563606-1: Upgrade ES Server Version 7.10.2 to 7.17.9 for Internal Elasticsearch

bhowd1:
### Changes
- upgraded elasticsearch server default version from 7.10.2 to 7.17.9
- upgraded elasticsearch default helm chart from 7.10.2 to 7.17.3 to support K8S version >=1.25

### Testing Scenario
- **Scenario 1: Fresh install on k8s 1.26 - ✅**

- **Scenario 2: Fresh install on k8s 1.23.7 - ✅**

- **Scenario 3: Upgrade existing chart to latest chart with changes on k8s 1.23.7**

  - **3.1 → with no change in values.yaml (es server version is same 7.10.2) ❌** (MAJOR IMPACT)
  - **3.2 → with changes in values.yaml (es server version as 7.17.9) ✅** (DATA IS STILL PERSISTED)
  
It solves #581


charts/backingservices/requirements.yaml
```diff
@@ -5,7 +5,7 @@
 # To disable deploying Elasticsearch in SRS, set the 'srs.srsStorage.provisionInternalESCluster' parameter in backingservices values.yaml to false.
 dependencies:
 - name: elasticsearch
-  version: "7.10.2"
+  version: "7.17.3"
```
reddy-srinivas, line 8: Why is this version 7.1.7.3 instead of 7.17.9? 
Is this intentional?

bhowd1: Hi Srini, `7.17.3 is the elastic helm chart version` which is used to spin up the elastic search cluster using helm.
Reference: https://github.com/pegasystems/pega-helm-charts/pull/647

kingakowalska1 (commented):
Hi Divyansh, 

I have a comment and a question. First, I think we should use the same ES version across all files (so the same one for values.yaml and requirements.yaml), even though we allow for different versions. I've noticed that you used 7.17.9 in the readme and values.yaml file, but 7.17.3 in requirements.yaml. I think we should settle on one example version to use everywhere.

Also, I have a question about the update process - I assume this is only necessary if you're updating both the Kubernetes version and ES version. Do you need to follow the process when updating the ES version even if you're on an earlier version of Kubernetes? I assume it's not needed if you create a new deployment with Kubernetes 1.25 and ES 7.17.9 from the start.

Thanks,
Kinga

bhowd1:
> Hi Divyansh,
> 
> I have a comment and a question. First, I think we should use the same ES version across all files (so the same one for values.yaml and requirements.yaml), even though we allow for different versions. I've noticed that you used 7.17.9 in the readme and values.yaml file, but 7.17.3 in requirements.yaml. I think we should settle on one example version to use everywhere.
> 
> Also, I have a question about the update process - I assume this is only necessary if you're updating both the Kubernetes version and ES version. Do you need to follow the process when updating the ES version even if you're on an earlier version of Kubernetes? I assume it's not needed if you create a new deployment with Kubernetes 1.25 and ES 7.17.9 from the start.
> 
> Thanks, Kinga

Hi @kingakowalska1 , Thanks for your review.
**Answer to your questions:**
- `I think we should use the same ES version across all files ` : The version 7.17.9 used in values.yaml file is the Elasticsearch server version whereas the version used in requirements.yaml file is the Elasticsearch helm chart version. They both are different. There are comments on top of the parameters too. Can you suggest if this comments doesn't help and we can handle this confusion in a better way?
- `I assume this is only necessary if you're updating both the Kubernetes version and ES version. Do you need to follow the process when updating the ES version even if you're on an earlier version of Kubernetes? I assume it's not needed if you create a new deployment with Kubernetes 1.25 and ES 7.17.9 from the start.`: This change is only necessary when the client wants to upgrade Kubernetes to version >=1.25. To do this upgrade they need to first upgrade the ES server to version 7.17.9 which supports k8s version >=1.25. And correct it's not needed if we create a new deployment with Kubernetes 1.25 and ES 7.17.9 from the start. 

Do let me know if you still have any questions.
Thanks

bhowd1:
Hi @kingakowalska1 ,
Thanks for your suggestions, I've updated the README file with your suggestions also I've modified it a bit since this change is only required for Internally provisioned Elasticsearch. Please check it in your free time.
Thanks

kingakowalska1 (approved):
Looks good! Thanks for the collaboration, @bhowd1 !

MadhuriArugula (approved):

kotamsetti (approved):
LGTM

Anand-Kotamsetti (approved):
LGTM

PR #658, opened 2023-10-25
Updated Hazelcast heap setting to derive from available RAM

vnihal72:
Updated Hazelcast heap setting to derive from available RAM

Changed the literal value to % so that it can be derived based on provided memory to the container.

Default from HZ : https://github.com/hazelcast/hazelcast-docker/blob/master/hazelcast-enterprise/Dockerfile#L16

### DIFF
--- a/charts/pega/charts/hazelcast/values.yaml
+++ b/charts/pega/charts/hazelcast/values.yaml
@@ -38,8 +38,8 @@ client:
   clusterName: "PRPC"
 # Server side settings for Hazelcast
 server:
-  java_opts: "-Xms820m -Xmx820m -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/opt/hazelcast/logs/heapdump.hprof
-  -XX:+UseParallelGC -Xlog:gc*,gc+phases=debug:file=/opt/hazelcast/logs/gc.log:time,pid,tags:filecount=5,filesize=3m"
+  java_opts: "-XX:MaxRAMPercentage=80.0 -XX:InitialRAMPercentage=80.0 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/opt/hazelcast/logs/heapdump.hprof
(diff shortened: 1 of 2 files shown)


MadhuriArugula (approved):

charts/pega/charts/hazelcast/values.yaml
```diff
@@ -38,8 +38,8 @@ client:
   clusterName: "PRPC"
 # Server side settings for Hazelcast
 server:
-  java_opts: "-Xms820m -Xmx820m -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/opt/hazelcast/logs/heapdump.hprof
-  -XX:+UseParallelGC -Xlog:gc*,gc+phases=debug:file=/opt/hazelcast/logs/gc.log:time,pid,tags:filecount=5,filesize=3m"
+  java_opts: "-XX:MaxRAMPercentage=80.0 -XX:InitialRAMPercentage=80.0 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/opt/hazelcast/logs/heapdump.hprof
+  -XX:+UseParallelGC -Xlog:gc*,gc+phases=debug:file=/opt/hazelcast/logs/gc.log:time,pid,tags:filecount=5,filesize=3m -XshowSettings:vm"
```
MadhuriArugula, line 42: How did we determine the metrics? Do we have any user story that showcases usage and metrics details? Please add appropriate details to PR description. 
Are the numbers specific to CloudK or CMC deployments?

Saurabh-16 (approved):

charts/pega/charts/hazelcast/values.yaml
vnihal72: Changed the literal value to % so that it can be derived based on provided memory to the container.

Default from HZ : https://github.com/hazelcast/hazelcast-docker/blob/master/hazelcast-enterprise/Dockerfile#L16

Its applied to both.

PR #669, opened 2023-11-20
Added truststore password input for TLS connection

anilkumargedela:
- Added truststore password input field with truststore path to support secure TLS connection between SRS and Elasticsearch


taz-pega-work (approved):
looks better ;)

kingakowalska1 (approved):
Look great!

charts/backingservices/charts/srs/README.md
```diff
@@ -76,6 +76,8 @@ make external-es-secrets NAMESPACE=pegabackingservices ELASTICSEARCH_VERSION=7.1
...
+| `certificateName`                       | Enter the tls certificate name.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
```
bhowd1, line 79: can we also add the default certificate name which is used if the certificateName value is not provided? (`elastic-certificates.p12`)

chandra-kalavakuri (approved):
lgtm

PR #677, opened 2023-12-08
BUG-801156 - Clarified messaging for Makefile commands for SRS deploy…

kingakowalska1:
…ment


charts/backingservices/charts/srs/README.md
```diff
@@ -43,8 +43,8 @@ You may enable the component of [Elasticsearch](https://github.com/helm/charts/t
...
-* To use an externally-provided Elasticsearch service with SRS, use the default `srs.enabled.true` parameter, update the `srs.srsStorage.provisionInternalESCluster` parameter in the [values.yaml](./values.yaml) to `false` and then provide connection details as documented below.
+* To use the internally-provided Elasticsearch service in the SRS cluster, use the default `srs.enabled.true` parameter and set the Elasticsearch version by updating the `elasticsearch.imageTag` parameter in the [values.yaml](./values.yaml) to match the `dependencies.version` parameter in the [requirements.yaml](../../requirements.yaml). This method streamlines the deployment process for development and testing environments, but it is not suitable for production environments, which require a fully external Elasticsearch cluster. Additionally, even though you deploy SRS and Elasticsearch together, Pega does not license the Elasticsearch cluster deployed using this method and does not maintain it as part of the Pega Platform support.
```
taz-pega-work, line 46: this part is vague and not really client-oriented>
Additionally, even though you deploy SRS and Elasticsearch together, Pega does not license the Elasticsearch cluster deployed using this method and does not maintain it as part of the Pega Platform support.

what do you think of saying it from a client perspective>
Additionally, when  you deploy SRS and Elasticsearch together, the deployment requires a connection to your existing Elasticsearch service maintained by your organization. Your organization must also maintain the license for your Elasticsearch service, since Pega does not include it in the SRS Docker image, nor will Pega provide it as part of Pega Platform container-based deployment support.

charts/backingservices/charts/srs/README.md
```diff
@@ -61,17 +61,15 @@ To deploy Pega Platform with the SRS backing service, the SRS helm chart require
 | `k8sProvider`                               | Specify your Kubernetes provider name. Supported values are [`eks`, `aks`, `minikube`, `gke`, `openshift`, `pks`].. 
 
 ### Enabling security between SRS and Elasticsearch
-To configure a secure connection between the SRS cluster and Elasticsearch, add the following the settings in your backingservices configuration file to reflect your organization's connectivity setup.
+Enabling a secure connection between SRS and your Elasticsearch cluster depends on the method you chose to deploy the Elasticsearch cluster.
```
taz-pega-work, line 64: I think it's better to stay conceptual here and not delve into clusters here, because the requirements are network-based. describing clusters feels like we should be talking about ports and nodes. Just my take, of course. The statement is fine as is...if you use sever, replace cluster with service throughout the readme

Enabling a secure connection between SRS and your Elasticsearch service depends on the method you chose to deploy it.

taz-pega-work (approved):
really nice improvements. I suggested a few edits but in general this is ready to go!

charts/backingservices/charts/srs/README.md
kingakowalska1: Hi Taz,
Thanks for your feedback! I'm not sure if your proposed changes are factually correct here: this paragraph is for a method of deploying an open source ES cluster TOGETHER with SRS for ease of deployment. The deployment of ES happens at the same time as SRS, so you do not connect to an already existing ES service, that is the point. The important part is to make it clear that Pega does not license this ES cluster even though the clients use our Helm charts to deploy it, that it is fully open source, and the Helm charts are there only to make the deployment process for testing environments easier. This is not recommended for production, though.

Furthermore, the messaging for this part was actually lifted verbatim from the docs.pega.com articles where I worked with the Search team to convey this in US-558611-1. I'd rather it stayed consistent across platforms.

charts/backingservices/charts/srs/README.md
kingakowalska1: This is a great suggestion, you're right! Thanks!

taz-pega-work (approved):
nice! thanks for updating it!

Anand-Kotamsetti (approved):
LGTM

PR #690, opened 2024-01-08
cleanup obsolete packed mode code

pega-roska:


khick77 (approved):

pega-roska:
We still probably need to do something about this, but this approach proved to not mesh very well with the underlying image.

PR #692, opened 2024-01-17
BUG-841263 (CMC only supported from 8.3+)

khick77:
Removed references to Pega Platform 8.2, as we only support 8.3+


kingakowalska1:
Hello,

Has the support policy changed, or was it always only supported for 8.3+?

khick77:
> Hello,
> 
> Has the support policy changed, or was it always only supported for 8.3+?

Apparently it was never supported, and inadvertently documented as if it were.

kingakowalska1 (approved):

misterdorito (approved):

PR #716, opened 2024-03-07
Create optional PDB for installer job (regardless of global.actions.execute)

zitikay:


charts/pega/charts/installer/templates/_pega-installer-job.tpl
```diff
@@ -22,6 +22,9 @@ metadata:
 {{ toYaml .root.Values.global.pegaJob.annotations | indent 4 }}
 {{- end }}{{- end }}
   labels:
+  {{- if (eq .root.Values.preventInterruption "true")   }}
+    pdb: "enabled"
+  {{- end }}
```
kishorv10, line 27: Would this be a generic selector? In the CMC, we are unsure of other deployments that Client may have along with Pega.

Probably selector like `app: installer` would be more specific.

charts/pega/charts/installer/values.yaml
```diff
@@ -34,6 +34,9 @@ bypassLoadEngineClasses: "false"
 bypassLoadAssembledClasses: "false"
 # If 'true', Helm will wait for the install or upgrade to finish, and only succeed if the job completes without error.
 waitForJobCompletion: "false"
+# If 'true', a PodDisruptionBudget will be created for the job's pod, preventing unexpected disruption
+#    ie from Cluster scaledown triggering node drain on node running job's pod
+preventInterruption: "false"
```
kishorv10, line 39: What are the use cases for not having it enabled by default?

charts/pega/charts/installer/templates/_pega-installer-job.tpl
zitikay: Yes the goal/aim here is to have a generic selector applicable for all global.actions.execute types. I have tested with CloudK install SCE, and verified PDB does not prevent job from cleaning up pod on its own, but *does* prevent node from draining while pod is running. If we didn't add new label, by default all job pods have label `job-name: "<job name>"`, but I don't want to conditionalize the selector based on the execute action (thus job name)

charts/pega/charts/installer/values.yaml
zitikay: Short answer, none that I can think of. I would think due to the nature of these jobs -- runs only once, can't really resume or retry -- that we'd want to prevent interruptions always. But was also wary of introducing the change in behavior, so instead chose to make it opt-in that we can set in PegaCloud, and optional for CMC

zitikay:
Since it was discussed on an internal webex channel, posting here for transparency:

I feel that we can enable PDB by default for install job.  Only condition I see for keeping it optional for install is min K8S version of 1.21, but end of life of 1.21 is more than one year back.

So, I think we can proceed making PDB default for install jobs.

zitikay:
So I'm open to putting in a check for k8s 1.21, or similar to the main pega chart pdb which changes the pdb apiVersion based on k8s version, but it's been EOL so long that we should be good introducing this change, potentially incrementing major helm chart version to indicate the lack of backwards compatibility with old k8s version.

zitikay:
Realized that pdb was being deployed unconditionally, even when action is "deploy" (no installer job needed, thus no pdb for job needed)

misterdorito (approved):

charts/pega/charts/installer/values.yaml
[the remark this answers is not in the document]
zitikay: I must've missed this comment. Based on above comments, removed the ability to set/unset this via values.yaml, now it's always on but only deployed if the job needs to be deployed

kishorv10 (approved):

PR #717, opened 2024-03-12
ISSUE-587: Added flag to disable ingress creation

kishorv10:
Added configuration option in values to control the ingress creation. Set `ingress.enabled` at tier level to true/false to control ingress creation. Removing `tier.ingress` portion from values also results in the same behaviour as enabled false.


MadhuriArugula:
Do we want to add Instructions regarding enable flag in (README)[https://github.com/pegasystems/pega-helm-charts/blob/master/charts/pega/README.md#ingress]

MadhuriArugula (approved):

Saurabh-16 (approved):

arvasrikanth (approved):

PR #718, opened 2024-03-12
Changing the setting to remove trailing /default so that it is not overrided when a different value is through DASS

Heshma:


misterdorito (approved):

APegaDavis:
I thought the purpose of adding this setting was to prevent clients from using different values in k8s envs?

APegaDavis (approved):

PR #732, opened 2024-04-09
make installer PDB a hook to avoid conflicts with upgrade jobs

APegaDavis:
In some scenarios install charts may be deployed twice, once for install and once for upgrade. This leads to a conflict on the installer PDB which is used for both.

This PR makes the PDB a hook to avoid that conflict.


zitikay (commented):
I was initially thinking we'd want pre-install/pre-upgrade hooks, to ensure we delete then create to avoid conflicts, but now actually leaning back towards post-* hooks because that just means once job is done we clean up PDB, and since job is done PDB isn't really needed anymore

khick77 (approved):

MadhuriArugula (approved):

Heshma (approved):

PR #739, opened 2024-04-25
Allow custom env in Installer job's pod

zitikay:
Provide a values.yaml entrypoint to add custom env vars for pod running installer job. Also refactored the environment config files to reduce duplication between install/upgrade, and always pass in ACTION env var the same way whether install or upgrade

### DIFF
--- a/charts/pega/charts/installer/templates/_pega-environment-config.tpl
+++ b/charts/pega/charts/installer/templates/_pega-environment-config.tpl
@@ -1,0 +1,69 @@
+{{- define "pega.installer.environment.config" -}}
+data:
+  # Temporary password for administrator@pega.com that is used to install Pega Platform and used to login after upgrading Pega Platform
+  ADMIN_PASSWORD: {{ .Values.adminPassword }}
+  # Database Type for installation
+  DB_TYPE: {{ .Values.global.jdbc.dbType }}
+  # JDBC URL of the DB where Pega is installed
+  JDBC_URL: {{ .Values.global.jdbc.url }}
+  # Class name of the DB's JDBC driver
+  JDBC_CLASS: {{ .Values.global.jdbc.driverClass }}
+{{- if .Values.global.jdbc.driverUri }}
+  # URI that the JDBC driver can be downloaded from
+  JDBC_DRIVER_URI: {{ .Values.global.jdbc.driverUri }}
+{{- end }}
+  # Rules schema of the Pega installation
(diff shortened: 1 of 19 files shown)


charts/pega/charts/installer/templates/_pega-environment-config.tpl
```diff
@@ -0,0 +1,59 @@
...
+  DATA_SCHEMA: {{ .Values.global.jdbc.dataSchema }}
+{{- if .Values.global.jdbc.customerDataSchema }}
+  # CustomerData schema of the Pega installation
+  CUSTOMERDATA_SCHEMA: {{ .Values.global.jdbc.customerDataSchema }}
+{{- end }}
+{{- if .Values.global.jdbc.connectionProperties }}
+  # JDBC custom connection properties
+  JDBC_CUSTOM_CONNECTION: {{ .Values.global.jdbc.connectionProperties }}
+{{- end }}
+  # Whether this is a Multitenant System ('true' if yes, 'false' if no)
```
pega-talba, line 27: This should go within the if statement so that it does not appear without the associated env var

charts/pega/charts/installer/templates/_pega-environment-config.tpl
```diff
@@ -0,0 +1,59 @@
+{{- define "pega.installer.environment.config" -}}
+data:
+  # Temporary password for administrator@pega.com that is used to install Pega Platform and used to login after upgrading Pega Platform
```
pega-talba, line 3:
```suggestion
  # Temporary password for administrator@pega.com that is used for the first login to Pega Platform on new installations.
```
Usually only relevant to installs. Only in exceptional circumstances is it needed for an upgrade. Probably not worth calling that out here as it gives the wrong impression of usage.

zitikay: Fair enough, was trying to capture the comments from both install + upgrade configmaps.
Upgrade: `  # Temporary password for administrator@pega.com that is used to login after upgrading Pega Platform`
But if that causes more confusion than it prevents, I can update this

pega-roska (approved):

charts/pega/charts/installer/templates/_pega-environment-config.tpl
zitikay: Also fair. This comes from copy pasting out of the install configmap which had it outside the if  
¯\\\_(ツ)\_/¯

pega-talba (approved):

Heshma (approved):

PR #769, opened 2024-06-05
BUG-868079 Add the option to configure keystore and truststore type for Cassandra connection

wonim2022:
…or cassandra client encryption


taz-pega-work (changes requested):
For all of the newly added comments, we need Cassandra to be capitalized so these comments have Cassandra consistently capitalized across all of the files. Otherwise, I think the wording looks pretty good.

wonim2022:
Thanks for the reivew, I capitalized all Cassandra occurrences in new comments

kingakowalska1:
Hi,

Could we please tweak the message a bit? My suggestion is as follows:

If required, provide the type of Cassandra truststore and keystore that you used to hold keys and certificates for client encryption. Available store types are JKS and PKCS12. Default is JKS.

Also, in terratest/src/test/pega/data/values_zdt_upgrade_http_non_default_port81.yaml, we should include both available types as well. I would suggest using the same message as in all the other files.

Thanks!
Kinga

wonim2022:
Tweaked the message like you proposed @kingakowalska1

taz-pega-work (approved):
changes look great. thanks for collaborating!

kingakowalska1 (approved):
Looks great!

baylp2022 (approved):

taz-pega-work:
completing this bug fix by merging the fix at the request of Paul Bayliss

PR #759, opened 2024-05-27
Issue-723: Exposed resource block as per pod spec to specify request and limits for Pega tiers

kishorv10:
Exposed new `resources` block in values.yaml as per Pod resource spec, moved the request & limit defaults of cpu & memory to the values.yaml.
Retained the defaults in code for backward compatibility for the deployments upgrading to newer versions but using old `values.yaml`.
Deprecated `cpuRequest, cpuLimit, memRequest & memLimit` configuration options in the favor of `resources` configuration.

PR #800, opened 2024-08-16
US-631955 changes to support deploying pods without cpu limit

NekhilKotha:


kishorv10:
@NekhilKotha The same requirement is addressed in PR #759 . With this change, values.yaml configuration can be used to set or not set CPU limits for Pega work loads. Please try, if that solves your usecase.

charts/pega/templates/_pega-deployment.tpl
```diff
@@ -225,9 +225,9 @@ spec:
           # Maximum CPU and Memory that the containers for {{ .name }} can use
           # Resources are configured through deprecated settings. Use .tier[].resources instead
           limits:
-          {{- if .node.cpuLimit }}
+          {{- if and .node.cpuLimit (ne (include "deployWithoutCPULimit" (dict "node" .node)) "true") }}
```
Saurabh-16, line 228: Can we simplify this in below sequence:-
If .node.cpuLimit is given, we always honour this.
else if we check for deploywithoutCPULimit
else we default it to 4 ?

Saurabh-16: Also how  CMC customers will be aware of this feature? Is there plan to provide some documentation around it.

NekhilKotha:
> @NekhilKotha The same requirement is addressed in PR #759 . With this change, values.yaml configuration can be used to set or not set CPU limits for Pega work loads. Please try, if that solves your usecase.

Hi @kishorv10, I have initially thought of using the same but we need this control of CPU Limit conditionally for controlplane Integration. So this is not possible to use that

kishorv10:
@NekhilKotha I believe with the latest developments, this PR may not be needed. Can you please confirm & close it accordingly?

PR #835, opened 2024-10-11
Replaced deploymentName with deployment.name in SRS chart

alemax22:
The key deploymentName in the SRS chart was replaced with the key deployment.name in order to be aligned to the style of all the other charts. The implementation maintains backward compatibility, in particular: the deploymentName (marked as deprecated) key has precedence over the new key deployment.name. A specific unit test was designed to assure this behaviour. 
All the other unit tests and the documentation were updated.


MadhuriArugula:
@pegasystems/searchservice Can you please review the changes

PR #845, opened 2024-11-06
Service Account configuration for constellation deployment

alemax22:
Add the option to configure a custom service account for the constellation pods.
This is needed because the constellation deployment needs a persistent volume claim, which needs to be shared by all the pods, the permissions to read/write on the shared storage can be controlled by a service account.
The documentation was updated to describe this new behaviour.


Saurabh-16:
@nikhilparmar86 :- Could you please take a look?

kishorv10 (approved):

alemax22:
Do we have any ETA for this item?

nikhilparmar86 (approved):

PR #856, opened 2024-12-04
US-648252 added job labels to the installer template

Evan1oconto:
Adding job labels to the installer template so there is the ability to add labels via the infinity service


Heshma (approved):

APegaDavis:
@Evan1oconto can you add a test to the [installer job tests](https://github.com/pegasystems/pega-helm-charts/blob/master/terratest/src/test/pega/pega-installer-job_test.go)?

APegaDavis (approved):

Issue #883, opened 2025-02-28
The OpenShift NetworkPolicy created for the backingservice when using Opensearch has missing labels causing the deployment to fail to connect to Opensearch


roller1187:
**Describe the bug**
When deploying an external Opensearch cluster on an OpenShift cluster, the NetworkPolicy created by the backingservices Helm chart blocks traffic between the srs-service deployment and the Opensearch cluster.

**To Reproduce**
1. Deploy and configure an external Opensearch cluster within OpenShift [reference](https://www.ibm.com/docs/en/product-master/12.0.0?topic=tasks-installing-opensearch-fix-pack-8-later)
2. Add PEGA Helm repo 
```bash
helm repo add pega https://pegasystems.github.io/pega-helm-charts
```
3. Modify the following values in backingservices.yaml accordingly:
```yaml
k8sProvider: 'openshift'
deploymentName: pega-search
srs.srsRuntime.srsImage: <REPLACE WITH THE LOCATION OF YOUR `pega-docker.downloads.pega.com/platform-services/search-n-reporting-service-os` IMAGE>
srs.srsStorage.provisionInternalESCluster: 'false'
srs.srsStorage.domain: <SET TO THE OPENSEARCH SERVICE ADDRESS IN OPENSHIFT>
srs.srsStorage.port: <SET TO THE OPENSEARCH SERVICE PORT IN OPENSHIFT>
srs.srsStorage.protocol: 'http'
srs.srsStorage.tls.enabled: 'false'
srs.srsStorage.authCredentials.username: 'admin'
srs.srsStorage.authCredentials.password: 'Pasword123!'
```
4. Run Helm chart for the backing services:
```bash
helm install backingservices pega/backingservices --namespace pega --values backingservices.yaml
```

**Expected behavior**
The srs-service "pega-search" pods should connect to the external Opensearch cluster previously deploy. Instead, the pods enter a CrashLoopBackOff state with the following stacktrace in the logs:
```
 __  __ _                                  _   
|  \/  (_) ___ _ __ ___  _ __   __ _ _   _| |_ 
| |\/| | |/ __| '__/ _ \| '_ \ / _` | | | | __|
| |  | | | (__| | | (_) | | | | (_| | |_| | |_ 
|_|  |_|_|\___|_|  \___/|_| |_|\__,_|\__,_|\__|
[...]

PR #884, opened 2025-02-28
Fixing NetworkPolicy for external Opensearch service on OpenShift deployments #883

roller1187:


ravitejamekapega (commented):
In case of opensearch, we recommend to disable default network policy(https://github.com/pegasystems/pega-helm-charts/blob/master/charts/backingservices/charts/srs/values.yaml#L53) and add new network policy for 'opensearch'

reddy-srinivas (changes requested):
Provisioning an internal OpenSearch cluster using Pega Helm charts is not supported; only external OpenSearch is supported as an option. Consequently, adding a podSelector for OpenSearch is unnecessary.

As for Elasticsearch, internal cluster provisioning is currently supported in non-production environments, which is why the podSelector option is available. However, discussions are ongoing about potentially discontinuing support for internal Elasticsearch provisioning.

So, you may withdraw this pr.
Thanks.

kishorv10:
Closing the PR as per the previous comments from @reddy-srinivas

PR #901, opened 2025-04-16
Explain the use of global.version field.

misterdorito:


charts/pega/values-large.yaml
```diff
@@ -17,6 +17,12 @@ global:
   actions:
     execute: "deploy"
 
+  # New versions of the Pega Platform require that certain libraries are included in the bootstrap classpath
```
pega-talba, line 20: "New" will become outdated eventually - let's call out the specific version.
```suggestion
  # Pega Platform versions 25.0 and above require that certain libraries are included in the bootstrap classpath
```

pega-talba (approved):

khick77 (approved):

Issue #904, opened 2025-04-26
Add support for custom entries in pod /etc/hosts


rob8714:
**Is your feature request related to a problem? Please describe.**
Sometimes custom entries need to be added to the /etc/hosts file in the pod

**Describe the solution you'd like**
Add support for including hostAliases in Pega deployment tier

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

misterdorito:
https://kubernetes.io/docs/tasks/network/customize-hosts-file-for-pods/

PR #905, opened 2025-04-26
ISSUE-904 - Add support for custom entries in pod /etc/hosts

rob8714:
PR for https://github.com/pegasystems/pega-helm-charts/issues/904

Note: Test was written with the help of one of the online Gen AI tools.


rob8714:
@kishorv10  The use case I used it for is to integrate a device that has a self signed certificate, and the hostname in the certificate could not be added to the DNS server being used.  Since the IP of the device was not in the SAN of the certificate, when sending the request over IP we were getting mismatch errors. In order not to have certificate issues due to a mismatch in hostname, we needed a way to map the IP address of the device to the hostname present in the certificate, which we did using the change submitted in this PR.  Once this was done, we did not face hostname mismatch issues and the integration worked fine.  This is something which is fairly simple to do on a VM by adding a hosts file entry, so the aim of the PR is to provide this functionality to containerized deployments as well.

kishorv10 (approved):
This support can be useful in dev use cases not recommended for production deployments

PR #915, opened 2025-05-16
Adding support for custom volumes in constellation chart 

### LINKED ISSUE #914: Add Custom volumes to constellation pods
**Is your feature request related to a problem? Please describe.**
In our kubernetes environment we cannot use persistent volume claims. However, we need to mount specific volumes that were previously created by our system admin.

**Describe the solution you'd like**
We would like to have custom volumes configuration like the one in the Pega chart 

**Describe alternatives you've considered**
NA

**Additional context**
NA

filippocasari:
This pull request solves fixes #914: we added a new optional key custom containing volumes and volumeMounts. We also documented the new feature and implemented related go tests.


SreecharanChallagundla:
this change is only when we want to add volumes or can this be by default.. if you make this by default that would be great help... our dev teams keeps on reporting when a constellation pod got recycled all their work is gone, if we add PVC then this issue will be resolved. As we know  when you add PVC we need to pick pod type as statefulsets. 

can we get this as default...!!!

PR #917, opened 2025-05-23
Only set replicas field on deployments when HPA is not enabled

APegaDavis:
Prior to this fix we set the `replicas` field on deployments to 1 or a user-specified value. When HPA is enabled and has a `minReplicas` value incosistent with the deployment's `replicas` value, it results in thrashing any time an update is made to the deployment.

For example, if `replicas` is set to `1` but HPA has `minReplicas` set to `4` the following is observed during an update to the deployment:

0. Deployment is running and HPA has set `replicas` to `4`
1. Deployment update is applied which resets `replicas` to `1`
2. The deployment-controller triggers and scales the replicaset down to 1 pod and extra pods are deleted
3. HPA triggers, sees the replicaset has less than `minReplicas` pods, and scales the deployment back up to `4`

[K8s documentation](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#migrating-deployments-and-statefulsets-to-horizontal-autoscaling) suggests excluding `replicas` field from the deployment manifest when using HPA. With this change we will skip setting `replicas` field on the deployment if `<tier>.hpa.enabled=true` or if hpa is not configured at all for the tier.


charts/pega/templates/_pega-deployment.tpl
```diff
@@ -28,8 +28,11 @@ metadata:
     app: {{ .name }} {{/* This is intentionally always the web name because that's what we call our "app" */}}
     component: Pega
 spec:
-  # Replicas specify the number of copies for {{ .name }}
+
```
misterdorito, line 31: I think that you should extract this into a named template that can be overridden in a subchart or else you may run into problems elsewhere.

misterdorito (approved):

pega-talba (approved):

PR #924, opened 2025-06-25
Fix logic for finding tier name with "WebUser" node type

ianlucas:
My Pega deployment had a single tier with name "pega", and when upgrading with zero-downtime action, the logic for inferring the Pega REST Server URL has the `contains` function arguments mixed up, so instead of resolving it to `http://pega-pega:80/prweb/PRRestService`, it would turn into `http://pega-web:80/prweb/PRRestService`.

#### Example `pega.yaml`

```yaml
# ...
  tier:
    - name: "pega"
      nodeType: "WebUser,BackgroundProcessing,Search"
      initialHeap: "2048m"
      maxHeap: "2048m"
      replicas: 1
      service:
        httpEnabled: true
        port: 80
        targetPort: 8080
        serviceType: "NodePort"
        tls:
          enabled: true
          port: 443
          targetPort: 8443
          traefik:
            enabled: false
# ...
```

**Expected:** `http://pega-pega:80/prweb/PRRestService`
**Actual:** `http://pega-web:80/prweb/PRRestService`

With this PR it should resolve properly and find `WebUser` inside `nodeType`'s value instead the other way around.

### DIFF
--- a/charts/pega/charts/installer/templates/_helpers.tpl
+++ b/charts/pega/charts/installer/templates/_helpers.tpl
@@ -223,7 +223,7 @@ currentFunctionPath=SYSIBM,SYSFUN,{{ include "resolvedDataSchema" . | upper }}
 {{- $webAppContextPath := "prweb" }}
 {{- range $index, $tier := .Values.global.tier }}
   {{- if hasKey $tier "nodeType" }}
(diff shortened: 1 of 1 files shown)


ianlucas:
@misterdorito do we still need this PR after #925's merge?

PR #928, opened 2025-07-03
US-687588: Updated elastic helm documentation to support ES 8.15.5 & 8.18.2

anilkumargedela:
Tested the search service with the ES 8.15.5 & 8.18.2 version, and it is working as expected. I also updated the documentation for using the ES 8.15.5 & 8.18.2 version.

### DIFF
--- a/charts/addons/Chart.yaml
+++ b/charts/addons/Chart.yaml
@@ -3,4 +3,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: A Helm chart for Kubernetes
 name: addons
-version: "3.28.4"
+version: "3.28.5"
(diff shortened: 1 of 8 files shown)


charts/backingservices/requirements.yaml
kingakowalska1, line 1: I think we might benefit from adding all versions of Elasticsearch that you can deploy using the 8.5.1 helm chart, so:

"For kubernetes version >=1.25 or Elasticsearch version 8.10.3, 8.15.1 or 8.18.2, use 8.5.1 for the Elasticsearch 'version' parameter below."

kingakowalska1 (changes requested):
Hi, 
It looks good, just one suggestion about the requirement.yaml file.

Kinga

kingakowalska1 (approved):
Looks good!

Saurabh-16 (approved):

PR #968, opened 2025-11-12
BUG-956927 : Pega helm installation fails when CUSTOM_JVM_ARGS is not set

pega-Abhinav:


Saurabh-16 (approved):

charts/pega/templates/_helpers.tpl
```diff
@@ -159,7 +159,7 @@
 
 # list of either external or internal cassandra nodes
 {{- define "cassandraNodes" }}
-  {{- if .Values.dds.externalNodes -}}
+  {{- if and (.Values.dds) (.Values.dds.externalNodes) -}}
```
kishorv10: @pega-Abhinav  why do these changes required for this PR/bug?

kishorv10 (approved):

PR #1010, opened 2026-04-28
Download jdbc libraries via init containers.

misterdorito:


charts/pega/charts/installer/charts/pegacommon/templates/_ic_lib_loader.tpl
```diff
@@ -0,0 +1,174 @@
...
+---
+# This template contains the script to download the JDBC driver from the specified location and copy it to the shared
+# volume. This script will be executed by the init container before starting the Pega application container.
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: {{ template "pegaLibDownloadScriptConfig" . }}
+  namespace: {{ .Release.Namespace }}
+data:
+  download-jdbc-lib.sh: |
```
pega-talba, line 90: Will this always be bash? We should probably call out bash in this case

charts/pega/charts/installer/charts/pegacommon/templates/_ic_lib_loader.tpl
```diff
@@ -0,0 +1,174 @@
...
+        else
+          echo "Using API key for artifactory authentication."
+          ca_auth="-H $CAAH:$CAAK"
+        fi
+      fi
+    fi
+
+    ca_cert=""
+    if [ "$(ls -A $art_root/cert/*)" ]; then
+      if [ "$(ls $art_root/cert/* | wc -l)" ]; then
```
pega-talba: Unless I am doing this wrong, `if [ "0" ]` resolves to true locally and this would always be true

terratest/src/test/pega/data/values_multidriver-with-cert.yaml
```diff
@@ -0,0 +1,17 @@
...
+global:
+
+  # Provide JDBC connection information to the Pega relational database
+  #   If you are installing or upgrading on IBM DB2, update the udb.conf file in the /charts/pega/charts/installer/config/udb directory with any additional connection properties.
+  jdbc:
+
+    #   For databases that use multiple JDBC driver files (such as DB2), specify comma separated values for 'driverUri'
+    driverUri: "http://driverhost/drivers/driver1.jar,http://driverhost/drivers/driver2.jar"
+
+  enableSSLVerification: true
```
pega-talba: We use `customArtifactory.enableSSLVerification` earlier - should this be inside the customArtifactory block?

charts/pega/charts/installer/charts/pegacommon/templates/_ic_lib_loader.tpl
```diff
@@ -0,0 +1,174 @@
...
+{{- $depName := printf "%s" (include "deploymentName" $) -}}
+{{- $depName -}}-{{- .chartType -}}-lib-download-script-config
+{{- end }}
+
+{{- define "jdbcLibVolume" }}
+{{- if .Values.global.downloadContainer }}
+{{- if .Values.global.downloadContainer.image }}
+- name: jdbc-lib-volume
+  emptyDir:
+    sizeLimit: 5Mi
```
pega-talba: Oracle's JDBC driver is 7.3 MB

rbogendoerfer: We should consider that there could be more than one single driver be downloaded here, e.g. drivers for additional databases of another type or drivers to access other systems like MQ, etc.
Hence I would suggest making the volume quite a bit bigger, e.g. 50MB.

khick77 (approved):

APegaDavis (approved):

pega-roska (approved):

pega-talba (approved):

PR #1023, opened 2026-07-01
T-3490 prlog4j2 traceid

stutibhushan29:
•Removed deploy-time static prlog4j2.xml; deploy now uses only charts/pega/config/deploy/prlog4j2.xml.tmpl.
•Kept correlation ID handling template-driven via PEGA_LOG_CORRELATION_ID_ENABLED (no changes to template logic itself).
•Updated tier ConfigMap template (pega-config.tpl) to expose only prlog4j2.xml.tmpl (removed prlog4j2.xml key).
•Updated deployment packaging (before_deploy.sh) to include only deploy/prlog4j2.xml.tmpl in deploy-config-.tgz.
•Updated XML validation script (validatexml.sh) to validate deploy prlog4j2.xml.tmpl instead of deploy prlog4j2.xml.
•Removed obsolete deploy fixture file terratest/src/test/pega/data/expectedInstallDeployPRlog4j2.xml.
•Updated Pega tier config tests to assert only prlog4j2.xml.tmpl content.
•Updated correlation ID tests to:
◦expect PEGA_LOG_CORRELATION_ID_ENABLED=false by default/when disabled
◦assert only prlog4j2.xml.tmpl key exists in tier ConfigMap
•Updated documentation references from prlog4j2.xml to prlog4j2.xml.tmpl (chart README + deployment guides for AKS/EKS/GKE/PKS/OpenShift).
•Full terratest suites passed: ./pega, ./addons, and ./backingservices.

PegaRule Logs:

traceIdEnabled : true

image

traceIdEnabled : false

image

### DIFF
--- a/charts/pega/config/deploy/prlog4j2.xml.tmpl
+++ b/charts/pega/config/deploy/prlog4j2.xml.tmpl
@@ -6,7 +6,7 @@
     </Console>
     <RollingRandomAccessFile  name="PEGA" fileName="${sys:pega.logdir}/PegaRULES.log" filePattern="${sys:pega.logdir}/PegaRULES-%d{MM-dd-yyyy}-%i.log.gz">
         <PatternLayout>
-            <Pattern>%d [%20.20t] [%10.10X{pegathread}] [%20.20X{tenantid}] {{ if eq .Env.PEGA_LOG_TRACE_ID_ENABLED "true" }}[%X{traceId}] {{ end }}[%20.20X{app}] (%30.30c{3}) %-5p %X{stack} %X{RequestorId} %X{userid} - %m%n</Pattern>
+            <Pattern>%d [%20.20t] [%10.10X{pegathread}] [%20.20X{tenantid}] {{ if eq .Env.PEGA_LOG_CORRELATION_ID_ENABLED "true" }}[%X{ext-correlation-id}] [%X{int-correlation-id}] {{ end }}[%20.20X{app}] (%30.30c{3}) %-5p %X{stack} %X{RequestorId} %X{userid} - %m%n</Pattern>
         </PatternLayout>
         <Filters>
             <!--Deny message logged under ALERT log level-->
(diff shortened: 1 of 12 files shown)


charts/pega/config/deploy/prlog4j2.xml.tmpl
```diff
@@ -0,0 +1,182 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="warn">
+<Appenders>
+    <Console name="CONSOLE" target="SYSTEM_OUT">
+        <LogStashJSONLayoutPega/>
+    </Console>
+    <RollingRandomAccessFile  name="PEGA" fileName="${sys:pega.logdir}/PegaRULES.log" filePattern="${sys:pega.logdir}/PegaRULES-%d{MM-dd-yyyy}-%i.log.gz">
+        <PatternLayout>
+            <Pattern>%d [%20.20t] [%10.10X{pegathread}] [%20.20X{tenantid}] {{ if eq .Env.PEGA_LOG_CORRELATION_ID_ENABLED "true" }}[%X{ext-correlation-id}] [%X{int-correlation-id}] {{ end }}[%20.20X{app}] (%30.30c{3}) %-5p %X{stack} %X{RequestorId} %X{userid} - %m%n</Pattern>
```
vargm, line 9: Conditionally adding/removing fields from the log pattern using:

```text
{{ if eq .Env.PEGA_LOG_CORRELATION_ID_ENABLED "true" }}
```

causes the actual log layout to change based on runtime configuration. This can create issues for tools that parse raw text log files because field positions shift depending on whether correlation ID logging is enabled or disabled.

For example:

With correlation IDs enabled:

```text
[tenantA] [EXT-123] [INT-456] [MyApp] ...
```

With correlation IDs disabled:

```text
[tenantA] [MyApp] ...
```

Any parser that relies on fixed field positions may require separate parsing logic for each variation of the pattern.

It can also make backward compatibility more difficult, as tooling that supports older log formats now has to handle multiple pattern variants.

Would it be better to keep the pattern stable and emit placeholder values when correlation IDs are disabled? For example:

```text
[tenantA] [-] [-] [MyApp] ...
```
or 

```text
[tenantA] [ ] [ ] [MyApp] ...
```

This preserves a consistent log structure regardless of configuration, simplifies downstream parsing, and reduces the maintenance burden for tools that need to support both current and older log formats.

Saurabh-16:
@stutibhushan29 :- Have we tested latest helm charts changes with old docker image? Also Latest docker changes with old helm chart? This is to maintain backward compatibility

APegaDavis:
@stutibhushan29 we'll also need to update the prlog4j2.xml to prlog4j2.xml.tmpl here: https://github.com/pegasystems/pega-helm-charts/blob/master/before_deploy.sh#L28

APegaDavis (approved):

pega-talba (approved):

APegaDavis (approved):

